PR #7, opened 2014-12-15
Allow integer arguments.

mshal:
This helps for things like the runId, which are ints in the task
structure.


jhford:
Looks good to me, thanks for the patch!

PR #9, opened 2015-01-15
Version 0.0.7

mshal:
Can you tag the repo as well? I don't think I can make a PR for a tag.


jhford:
And uploaded to Pypi

https://pypi.python.org/pypi/taskcluster/0.0.7

PR #10, opened 2015-01-21
Bug 1124110 - log HTTP traffic in debug output for communication with API endpoints

petemoore:
This change adds logging of http headers and payloads (both request and response) when debugging is enabled (via DEBUG_TASKCLUSTER_CLIENT env var) in taskcluster-client.py library.

This is to aid troubleshooting in the case an API call fails, but it is not understood the reasons why, or what the response was from the API endpoint.

### DIFF
--- a/taskcluster/utils.py
+++ b/taskcluster/utils.py
@@ -125,8 +125,17 @@ def makeHttpRequest(method, url, payload, headers, retries=MAX_RETRIES):
 
 
 def makeSingleHttpRequest(method, url, payload, headers):
-  log.debug('Making a %s request to %s', method, url)
-  return requests.request(method.upper(), url, data=payload, headers=headers)
+  log.debug('Making a %s request to %s', method.upper(), url)
+  log.debug('Sending HTTP request headers:')
+  log.debug(headers)
+  log.debug('Sending HTTP request payload:')
+  log.debug(payload)
+  response = requests.request(method.upper(), url, data=payload, headers=headers)
+  log.debug('Received HTTP response headers:')
+  log.debug(response.headers)
+  log.debug('Received HTTP response body:')
+  log.debug(response.text)
+  return response
 
 
 def putFile(filename, url, contentType):


jhford:
Looks good, with a couple of changes, we can merge this.  Thanks for the patch!

petemoore:
Hi John,
Thanks for your feedback!
The reason I logged method.upper() is because that is what is sent down to requests library (I didn't change this - see https://github.com/taskcluster/taskcluster-client.py/blob/e916ae6fa5e392bac824abe9ee479536c08430d7/taskcluster/utils.py#L129).
I'm wondering if we only log first 1024 characters of data from the http request, whether we might miss important information in the debug logs due to it being truncated. I also see that we don't want the logs to blow up (I guess this is why you'd prefer to truncate). If this is the case, another option might be instead of logging this in utils.py, logging instead the output in taskcluster/client.py in case of failure only, e.g. here:
https://github.com/taskcluster/taskcluster-client.py/blob/e916ae6fa5e392bac824abe9ee479536c08430d7/taskcluster/client.py#L350
The advantage of this approach is that we would only see this high level of detail in the case of a failure, and in such a case it might make sense to log the full headers and body in raw form, so it is known exactly what was sent over the wire and back.
I'm happy to make a patch for this alternative implementation if you like. Let me know what you think! The reason I like to log it in raw form is to have visibility of the exact data sent, in case of problems during encoding etc.
Pete

petemoore:
... and I just realised the request headers is of course not raw text, but a dictionary - so I get now what you mean about encoding in json, rather than logging as a python dictionary as in above. Alternatively if the requests library gives us access to the actual bytes it sent and received for the headers, that might be another option for outputting raw headers, so that we don't need to be concerned with how the requests library transformed the dictionary to an actual byte stream, but this maybe is overkill and json is good enough. =)

jhford:
My concern here is that modifying the string in two places is a source for confusion in the future.  Let's create a variable at the beginning of the method and assign method.upper() to it and use that in the body.

My concern is that if a server spits out >1mb of data as a response that we cause logs and terminals to go berserk.  Useless data because of truncation is just as useless if your screen has 1000lines of WTF-8 from a binary response.  In the case that you need to examine anywhere near a megabyte of data, I'd suggest that writing that to a file is necessary for any real analysis.  I would assume that the majority of uses for this are just seeing what the error message from the server is, which should be well under 1024.

Thinking more about it, for non-500 errors we should probably try to parse the payload and set the Exception message to be the server's response instead of boilerplate code.  That should be another bug.  If we do this, we remove the case where a failure happens in production but want to know what the error was.  Let's leave all headers and all payloads enabled with debug mode on and avoid sprouting more configuration flags :)

As for raw bytestreams, I feel like I'd much rather read a single level JSON dictionary than a really long string with lots of linebreaks, etc.  Anything more advanced feels like a problem for the debugger and not printf debugging.  Of course we should only be doing this JSON serialization if debugging is on.  I think python logging lets you see what the effective logging level for a log handler is, which lets us conditionally do all this extra logging.

jhford:
I ended up doing this myself

petemoore:
Thanks John!

PR #11, opened 2015-02-13
fix up logging

catlee:
You get double logging right now if you enable DEBUG_TASKCLUSTER_CLIENT because both modules add a StreamHandler to logging.

It's generally left up to the application code to configure logging, and modules should just log messages. As a compromise I've left the modules to configure logging in a safe way if the DEBUG_TASKCLUSTER_CLIENT var is set.


jhford:
@catlee I've modified the logging behaviour and I've removed the dependency on node for generating the API reference and added some more documentation to the README.md.

If there are still issues, please feel free to open a new pull request.

PR #13, opened 2015-02-19
Fix various retry bugs, or hopefully just some of them

jonasfj:
Various small fixes... 

 * Bug 1134369
 * Bug 1133876
 * Bug 1133856
 * Bug 1133048
 * Bug 1133020

Please let me know if you have any questions... I'm not sure tests runs correctly, I keep seeing a bug in my tests... but I think it's just obscure setup...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ update-readme: $(VENV)/bin/python
 
 $(VENV)/bin/python:
 	[ -d $(VENV) ] || $(PYTHON) -m virtualenv $(VENV) || virtualenv $(VENV)
-	$(VENV)/bin/pip install setuptools
+	$(VENV)/bin/pip install --upgrade setuptools
 	$(VENV)/bin/python devDep.py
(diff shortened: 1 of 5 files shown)


jonasfj:
I got the same error locally as on travis:
```
nosetests
......../opt/python/2.7.9/lib/python2.7/unittest/case.py:340: RuntimeWarning: TestResult has no addExpectedFailure method, reporting as passes
  RuntimeWarning)
```
I have no idea why... @jhford do you have a clue?

jhford:
I have seen that failure before and I can't remember what caused it.

jhford:
The error is because of the use of `@unittest.expectedFailure` for the temp credentials tests.  I'm not sure why that doesn't always show up as an error.  The support for that decorator was merged into nose.py this month, so maybe this won't be an issue for long.

jonasfj:
@jhford, can you make sure this get pushed to pypi ?

PR #17, opened 2015-07-14
Return response obj for downloading artifacts

askeing:
- ex: getLatestArtifact() of Queue API


jhford:
The only concern that I have is for files which are binary, think a firefox dmg, we might want to force the response to be a base64 encoded version of the response.

jhford:
I don't know where my mind is today... this is the client object, so we're not reserialising the response object, which means that base64 encoding would be pointless.

Merging now.

PR #18, opened 2015-07-15
Support steaming download.

askeing:
- remove response.text in util (fix the UnicodeEncodeError when downloading files)
- run response.json only if the headers['content-type'] is json format

### DIFF
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -426,7 +426,11 @@ class BaseClient(object):
 
       # Try to load JSON
       try:
-        return response.json()
+        content_type = response.headers.get('content-type')
+        if content_type is not None and 'json' in content_type.lower():
+            return response.json()
+        else:
+            return {"response": response}
       except ValueError:
         return {"response": response}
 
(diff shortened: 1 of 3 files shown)


jhford:
The unit tests don't pass with this patch.  Can you make sure that they pass?  Without credentials, you can test everything with `NO_TESTS_OVER_WIRE=1` in your environment.

jhford:
It looks like the`ObjWithDotJson` needs to be updated...

you should be able to run the tests with `make tests`

askeing:
@jhford , I updated the object, thanks for your help :)

jhford:
0.0.21 is on pypi with this change

jonasfj:
@jhford, I think this is bad. It gives and inconsistent interface, doesn't handle retries and can't be reimplemented in the go client.

Note, some artifacts may have content-type JSON.

I suggest we discuss a better approach in bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1184209

It's probably something we can easily add to the reference format.

PR #19, opened 2015-07-16
Remove the response.text

askeing:
- if the response is artifact, then it will fail


jonasfj:
@jhford, this looks sane to me... Thoughts?

@askeing, I don't think using the API methods to download artifacts is a good idea.
See my comment in previous PR, I filed a bug for how we can make an API to stream down binary responses. But right now the best way to access artifacts is to build a signed url for API method.

jhford:
this will land as part of another branch

jhford:
This has been addressed as part of the 1.0.0 series of releases.

PR #20, opened 2015-07-28
Add support for encryption and stable slugId

rail:
This adds support for encrypted environment variables and stable slugIds. We've been using these in Funsize for a while now, so they can be defined as stable.


rail:
I didn't add any verification because the time format used in the message may change soon from unixtime to ISO time. It won't be hard to implement if someone starts working on python worker implementation.

jhford:
Thanks for the PR.  This will be included in the next release, ping me if you need it before I do the next release!

PR #22, opened 2015-08-29
Scope match method

jonasfj:
Hi, welcome back... :)

Just in time to take a look a this... I didn't know python `for ... else` was so nice...

If you could give a hit on how to write tests I'll happily crank out a few...


bhearsum:
@jonasfj, tests should be pretty simple, something like:
import unittest

from taskcluster.utils import scope_match

class TestScopeMatch(unittest.TestCase):
  def testHasExactScope(self):
    self.assertTrue(scope_match([["foo"]], ["foo"]))

  def testHasGlobbedScope(self):
    self.assertTrue(scope_match([["foo"]], ["fo*"]))

etc.

jonasfj:
wow I had to fight the linter and type a lot (I couldn't figure out how to generate test cases easily like dustin did in [tc-base](https://github.com/taskcluster/taskcluster-base/blob/a1e0ab362ceb8996fa463d0b9682e20fdcc84df0/test/scopematch_test.js#L23-L68))..

 - but here the test cases are :)

Lots of lovely test cases...!

jhford:
Looks good to me!

I don't know the way to get a lot of named test cases in python's unittest library, since they're just methods on an object.  We could probably futz around with generating the functions, naming them, then creating the unit test subclass dynamically with `type`.  That's huge overkill though.

The other option would just be to have a list of the args to the assertSc... method and have a single test that iterates over the test cases.

e.g.
```python
cases = [
  [['a'], ['b'], 'exception', 'description of this test case'],
]

def test_all(self):
  for case in cases:
    self.assertScopeMatches(*case)
``` 

r+ as is, and I'd also be ok with writing the test cases out as a static list.

gregarndt:
maybe a case for py.test's parametrize decorator?

jonasfj:
This isn't that bad... once written we can fix it if we ever refactor how scopes are validated :)

Note, using multi caret editor it's not super bad... but let's call it a fun challenge :)

jonasfj:
@jhford, can you publish a new version to pypy?

jhford:
0.0.25 released

PR #23, opened 2015-09-03
Ordered docs

jhford:
Similar to the js client's https://github.com/taskcluster/taskcluster-client.py/commit/334160eaa8778e4a5410b70cf58e894b80db6c99 the methods in the README.md file are now sorted by api name or exchange title.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,50 +81,123 @@ figuring out how to configure the python logging module but do want debug messag
 
 <!-- START OF GENERATED DOCS -->
 
-### Methods in `taskcluster.Index`
+### Methods in `taskcluster.Auth`
 ```python
-// Create Index client instance
+// Create Auth client instance
 import taskcluster
-index = taskcluster.Index(options)
+auth = taskcluster.Auth(options)
 ```
-#### Get Artifact From Indexed Task
- * `index.findArtifactFromTask(namespace, name) -> None`
- * `index.findArtifactFromTask(namespace='value', name='value') -> None`
+#### Get Client Authorized Scopes
+ * `auth.scopes(clientId) -> result`
+ * `auth.scopes(clientId='value') -> result`
 
-#### Find Indexed Task
- * `index.findTask(namespace) -> result`
- * `index.findTask(namespace='value') -> result`
+#### Get Client Credentials
+ * `auth.getCredentials(clientId) -> result`
+ * `auth.getCredentials(clientId='value') -> result`
 
-#### Insert Task into Index
- * `index.insertTask(namespace, payload) -> result`
- * `index.insertTask(payload, namespace='value') -> result`
+#### Get Client Information
+ * `auth.client(clientId) -> result`
+ * `auth.client(clientId='value') -> result`
 
-#### List Namespaces
- * `index.listNamespaces(namespace, payload) -> result`
- * `index.listNamespaces(payload, namespace='value') -> result`
+#### Create Client
+ * `auth.createClient(clientId, payload) -> result`
+ * `auth.createClient(payload, clientId='value') -> result`
 
-#### List Tasks
- * `index.listTasks(namespace, payload) -> result`
- * `index.listTasks(payload, namespace='value') -> result`
+#### Modify Client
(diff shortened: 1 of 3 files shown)


petemoore:
We can do this, I just wanted to highlight that in the js client, we order the apis (i.e. they are in the following order: auth, awsprovisioner, index, purgecache, queue, scheduler) but inside the docs for each api, the methods are shown in the order they appear in the reference schema.

What is unfortunate about our existing design, is that the json reference schema provides the method definitions as a json object in the reference schemas, yet [_a json object is an unordered set of name/value pairs_](http://json.org/).

In other words, as soon as you consume the reference schema, depending on the json library you use, you might disrupt the serialised order of the entries.

For this reason, if the original ordering was important, we probably shouldn't have used json to publish the references, or we shouldn't have used a json object to contain the methods, or we should have provided extra metadata to define the ordering.

I personally don't have a preference either way, as long as the ordering is deterministic (for example in the godocs of the go taskcluster client, the method names are listed in lexicographical order, due to the godoc design).

At the moment in the python client README.md, the apis/exchanges are listed in the order:

* Methods in `taskcluster.Index`
* Methods in `taskcluster.PurgeCache`
* Exchanges in `taskcluster.AwsProvisionerEvents`
* Methods in `taskcluster.Auth`
* Methods in `taskcluster.Queue`
* Exchanges in `taskcluster.PurgeCacheEvents`
* Exchanges in `taskcluster.QueueEvents`
* Methods in `taskcluster.Scheduler`
* Methods in `taskcluster.AwsProvisioner`
* Exchanges in `taskcluster.SchedulerEvents`

It might be nice to sort these [here](https://github.com/taskcluster/taskcluster-client.py/blob/37b4881d4482d229f8c98ae6da1620cac4bc5dff/genDocs.py#L82) so that they appear in this order:

* Methods in `taskcluster.Auth`
* Methods in `taskcluster.AwsProvisioner`
* Methods in `taskcluster.Index`
* Methods in `taskcluster.PurgeCache`
* Methods in `taskcluster.Queue`
* Methods in `taskcluster.Scheduler`
* Exchanges in `taskcluster.AwsProvisionerEvents`
* Exchanges in `taskcluster.PurgeCacheEvents`
* Exchanges in `taskcluster.QueueEvents`
* Exchanges in `taskcluster.SchedulerEvents`

Thanks John!

jonasfj:
>What is unfortunate about our existing design, is that the json reference schema provides the method definitions as a json object

@petemoore, just to be clear the only unordered thing is the URL entries in manifest.json

The list of methods within each is ordered...
If we cared about order in the manifest we shouldn't have used an object and instead use an array of objects...

Either way we should definitely use Json, not XML he he :)

IMO none of this matters there are bigger fish and manifest.json is still a temporary solution... I'm not sure what the permanent one is, maybe it is manifest, but I suspect not due to consistency issues in s3...
Anyways that's the whole docs, worker, usage.js, references discussion we have to figure out in Berlin... Wow we'll need beer to get through that :)

petemoore:
> > What is unfortunate about our existing design, is that the json reference schema provides the method definitions as a json object
> @petemoore, just to be clear the only unordered thing is the URL entries in manifest.json

My mistake - I've updated my comment.

> The list of methods within each is ordered...

> > If we cared about order in the manifest we shouldn't have used an object and instead use an array of objects...
> Either way we should definitely use Json, not XML he he :)

Agreed!

> IMO none of this matters there are bigger fish and manifest.json is still a temporary solution... I'm not sure what the permanent one is, maybe it is manifest, but I suspect not due to consistency issues in s3...
> Anyways that's the whole docs, worker, usage.js, references discussion we have to figure out in Berlin... Wow we'll need beer to get through that :)

OK, we'll add it to the agenda! Agreed the ordering is a very small matter compared to other matters we have to resolve at the moment.

jhford:
Ok!  I'll just order the manifest.json.  I thought it was a little odd to be alphabetizing all the methods and exchanges, guess my post-pto haze tripped me into not reading the part where it was only the manifest :)

petemoore:
:+1:

PR #24, opened 2015-09-03
Bug 1194767 - use 'nice' slugids from slugid.py package

petemoore:
Regarding the tests, now that the new slugid library has [its own tests](https://github.com/taskcluster/slugid.py/blob/master/test.py), there is an argument we could remove the slug tests from this library. However maybe it is also nice to keep them as an extra safety guarantee. The current tests in this library stub the `uuid.uuid4()` function since previously the slugid generation process in this taskcluster client was using `uuid.uuid4()`. The new [slugid client](https://pypi.python.org/pypi/slugid) is also using `uuid.uuid4()`, but since now this is implemented in an external library, this could change, so if we keep the tests we also now making assumptions about the other library's implementation. This might be another reason to remove the tests.

I don't mind either way, let me know what you think.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,14 +17,14 @@ tests_require = [
   'subprocess32==3.2.6',
   'psutil==2.1.3',
   'hypothesis',
-  'pgpy',
-  'slugid'
+  'pgpy'
 ]
 
 install_requires = [
   'requests>=2.4.3,<=2.7.0',
   'PyHawk_with_a_single_extra_commit==0.1.5',
   #  'PyHawk==0.1.4',
+  'slugid',
 ]
 
 if __name__ == '__main__':
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,8 @@ tests_require = [
   'subprocess32==3.2.6',
   'psutil==2.1.3',
   'hypothesis',
-  'pgpy'
+  'pgpy',
+  'slugid'
 ]
 
 install_requires = [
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -17,7 +17,8 @@
   'subprocess32==3.2.6',
   'psutil==2.1.3',
   'hypothesis',
-  'pgpy'
+  'pgpy',
+  'slugid'
```
rail: Shouldn't this go into install_requires?

taskcluster/utils.py
```diff
@@ -1,11 +1,11 @@
 import re
 import json
 import datetime
-import uuid
 import base64
 import logging
 import os
 import requests
+import slugid as slugid_
```
rail: I don't think you need to use "as slugid_" here. It doesn't collide with other variables. Just wanted it have it more distinguishable from slugId()?

rail:
Otherwise lgtm!

taskcluster/utils.py
petemoore: Ah yes, my mistake! The function defined below is `slugId` not `slugid` so there was no collision. Good spot. :+1:

setup.py
petemoore: Whoooooooooooops :grimacing:

jhford:
Looks good to me!  feel free to merge whenever you like

PR #26, opened 2015-09-07
Some tweaks to the release script

petemoore:
Added some extra safety checks to release process.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -114,5 +114,5 @@ git push "${OFFICIAL_GIT_REPO}" "+refs/tags/$VERSION:refs/tags/$VERSION" "+refs/
 python setup.py sdist
 # use twine to upload to protect pypi credentials (python setup.py upload can
 # send creds in cleartext)
-pip install twine
+source env-python/bin/activate
 twine upload -s dist/*
--- a/release.sh
+++ b/release.sh
@@ -17,6 +17,10 @@ set -e
 # Phase 1: Testing things
 
 # VERSION should have form x.y.z where x, y, z are positive integers
+# Note that \(0\|[1-9][0-9]*\) means that the only allowed number to
+# begin with 0 is the number 0 itself. All other numbers begin with
+# [1-9]. Also this regex requires at least one character, so '' does
+# not match.
 if ! echo "${VERSION}" | grep -q '^\(0\|[1-9][0-9]*\)\.\(0\|[1-9][0-9]*\)\.\(0\|[1-9][0-9]*\)$'; then
   echo "VERSION = '${VERSION}'"
   echo "Please set it to a release version, of the form x.y.z where x, y, z are positive integers"
@@ -38,6 +42,8 @@ if [ -n "$modified" ]; then
   echo "There are changes in the local tree.  This probably means"
   echo "you'll do something unintentional.  For safety's sake, please"
   echo 'revert or stash them!'
+  echo
+  git status
   exit 66
 fi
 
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -6,33 +6,60 @@
...
+# exit in case of bad exit code
 set -e
 
 # Phase 1: Testing things
-if [ -z $VERSION ] ; then
-  echo "You must specify a version using \$VERSION"
-  exit 1
+
+# VERSION should have form x.y.z where x, y, z are positive integers
+if ! echo "${VERSION}" | grep -q '^\(0\|[1-9][0-9]*\)\.\(0\|[1-9][0-9]*\)\.\(0\|[1-9][0-9]*\)$'; then
```
jhford: i'm not a regex expert, but doesn't `0\|[1-9][0-9]*` equate to `[0-9]+`?  My understanding is that the pipe is an or operator that works on the next token, in this case [1-9].  If they're both [0-9], then using + with one instead of * would acheive the same result.  I'm probably missing something here, it's been a while since i've had to write a big regex

release.sh
```diff
@@ -6,33 +6,60 @@
...
+  echo "The remote taskcluster repo is on commit ${remoteMasterSha}."
+  echo "Make sure to git push/pull so that they both point to the same commit."
+  exit 67
 fi
 
-set -x;
+set -x
+
+# Make sure that build environment is clean
+git clean -fdx
```
jhford: I really don't like this.  I don't want the release script to automatically clean up my repo as I often have little hacky scripts, logs and the sort which don't affect the build.  I understand why we'd want to ensure a clean environment, so could we instead do something where we exit if the build environment is not clean instead of automatically cleaning?

release.sh
```diff
@@ -43,7 +70,12 @@ make
 # Phase 2: Modifying and pushing
 
 # Version number
-sed -i "s,^VERSION=.*$,VERSION=\'$VERSION\',g" setup.py
+# Avoid sed -i to be mac compatible
```
jhford: I always forget that I install the GNU coreutils on my mac without the g prefix!

release.sh
```diff
@@ -62,5 +94,9 @@ fi
 # only file which is changing
 git commit -m "Version $VERSION" $commitFiles
 git tag "$VERSION"
-git push github.com:taskcluster/taskcluster-client.py master "$VERSION"
-python setup.py sdist upload
+git push "${OFFICIAL_GIT_REPO}" "$VERSION:$VERSION"
```
jhford: we still need to push the `master` reference or else we'll never have the changes from this scripts commits to setup.py and readme.md on the master branch of the repo.

release.sh
petemoore: It is a messy regex, hey! The reason I used this form is that the only number that can start with 0 is 0 itself, i.e. we don't want a version like 04.02.30 but rather 4.2.30. So this means either a 0 by itself, or 1 of [1-9] followed by 0 or more of [0-9]. This way the only number that starts with a 0 is 0, and we also ensure that each number has at least one character (so the empty string does not match).

release.sh
petemoore: Funny, I was thinking the same thing after sending it, and I was contemplating pushing another commit before you review. I'm more than happy to do that. The only thing this should clean is stuff that is listed in the .gitignore, as everything else should be found with the git status command above. I'll make sure it exits if it finds a change there, with the details, and the suggestion to run git clean -fdx to fix it so a user can copy/paste if they decide it is a good idea.

release.sh
petemoore: I should probably do that too!

release.sh
petemoore: Very good point, I'll add that to the push. Nice spot.

jhford:
I'm going to merge this, but in the future we might need to change how it forces us to only do a release off HEAD of master branch if we need to do an emergency release.

Thanks!

PR #29, opened 2015-10-16
Quick hack at bug 1215713

jonasfj:
Not sure this is right... it's just a quick hack auto testing will hopefully validate...

### DIFF
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -14,6 +14,7 @@ import datetime
 import calendar
 import requests
 import time
+import urllib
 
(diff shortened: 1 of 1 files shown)


jhford:
I can't log into bugzilla right now (yay 2fa) and I can't seem to run the tests on my laptop because of a weird openssl issue.... I'll fix this tomorrow

jonasfj:
He he tomorrow is fine... It is sunday today (or I have seriously over slept!!!)..

Thanks for picking up it up...

Issue #15, opened 2015-03-04
Error when credentials are unicode strings


catlee:
If your taskcluster credentials are stored in unicode strings (which will be common if you're loading them with json.load()), you end up with failures like this:

```python
  File "/home/catlee/mozilla/taskcluster-client.py/taskcluster/client.py", line 442, in apiCall
    return self._makeApiCall(e, *args, **kwargs)
  File "/home/catlee/mozilla/taskcluster-client.py/taskcluster/client.py", line 224, in _makeApiCall
    return self._makeHttpRequest(entry['method'], route, payload)
  File "/home/catlee/mozilla/taskcluster-client.py/taskcluster/client.py", line 350, in _makeHttpRequest
    header = hawk.client.header(url, method, hawkOpts)
  File "/home/catlee/.virtualenvs/bbb/local/lib/python2.7/site-packages/hawk/client.py", line 126, in header
    mac = hcrypto.calculate_mac('header', cred, artifacts)
  File "/home/catlee/.virtualenvs/bbb/local/lib/python2.7/site-packages/hawk/hcrypto.py", line 35, in calculate_mac
    result = hmac.new(credentials['key'], normalized, digestmod)
  File "/usr/lib/python2.7/hmac.py", line 136, in new
    return HMAC(key, msg, digestmod)
  File "/usr/lib/python2.7/hmac.py", line 75, in __init__
    self.outer.update(key.translate(trans_5C))
TypeError: character mapping must return integer, None or unicode
```

The problem is that the accessToken is being used as the hmac key, and hmac.new() doesn't support unicode keys at all.

We should probably encode the credentials strings to ascii. It's arguable that this should be done in the hawk library itself rather than tc-client.py. Either would WFM!

jhford:
I'm sure that it's possible to do this in the PyHawk Library and probably the better option, but until someone is able to actually make releases of the PyHawk library, I suspect that we'll have much better luck doing this in the client.

catlee:
Maybe we could try the mohawk library, it seems to be more active.

jhford:
I looked into using mohawk, but I found that it lacks the bewit generation that we need for signed URLS.
```
TODO
    Implement bewit. The bewit URI scheme is not implemented at this time. 
```

jhford:
@catlee does this resolve the issue for you?

PR #31, opened 2015-11-02
quote / to %2F in an argument

djmitche:
I recall some discussion of this before, but I don't remember if it applied to the Python client.  Without this hack, I can't fetch roles with `/` in the name.

### DIFF
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -304,7 +304,7 @@ class BaseClient(object):
       toReplace = "<%s>" % arg
       if toReplace not in route:
         raise exceptions.TaskclusterFailure('Arg %s not found in route for %s' % (arg, entry['name']))
-      val = urllib.quote(str(val).encode("utf-8"))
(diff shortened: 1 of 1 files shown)


jonasfj:
Hmm... PR #15 should have solved this... Are you sure that has made it to pypi?

If not what other escaping bugs do we have...

djmitche:
No, #15 is unrelated -- that's about unicode strings and hawk

jonasfj:
Sorry, ment pr #29, https://github.com/taskcluster/taskcluster-client.py/pull/29

djmitche:
No, that didn't fix it because `urllib.quote` doesn't quote `/` by default.  That's what the additional argument is ("safe", with default value `"/"`).

jonasfj:
Ahh... Nice...

@jhford, seems like is ready to ship... What do you think...

jhford:
lgtm

PR #33, opened 2015-12-17
Added authenticate method for using login.taskcluster.net

jonasfj:
This is lovely crazy and awesome... mozregression should probably used this, it would save them a lot of pain (I hope).

It needs docs in `README.md` as I don't really know how the `utils.py` thing is exposed...

Or we should just generate readthedocs things... for this...

### DIFF
--- a/taskcluster/utils.py
+++ b/taskcluster/utils.py
@@ -7,6 +7,11 @@ import os
 import requests
 import slugid
 import time
+import webbrowser
+import urlparse
+import BaseHTTPServer
+import urllib
+import sys
 
 MAX_RETRIES = 5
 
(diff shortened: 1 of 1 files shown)


nhirata:
Travis push failed...

jhford:
this is great!  As an extra method of getting credentials, I think this will make life a lot easier when writing simple scripts.  I wonder if there's value in having an option in the script to automatically read from the environment variable and only open the login.taskcluster.net stuff when we really need to.

We should also support a `~/.tc-credentials` file and offer the option after this method to write it out there, similar to how Travis and Hub (github's cli) do.  If it also read in the environment variables to possibly short circuit the function, we could have people write scripts locally that don't need to be changed for production potentially, and have production scripts easy to run locally...

Anyway, r+ as is, but also happy to review or help write the credential storage stuff.

jonasfj:
> We should also support a ~/.tc-credentials file and offer the option after this method to write it out there, similar to how Travis and Hub (github's cli) do. If it also read in the environment variables to possibly short circuit the function, we could have people write scripts locally that don't need to be changed for production potentially, and have production scripts easy to run locally...

This would be awesome... Please note that the correct location is `~/.config/taskcluster-credentials.json` or something like that, see XDG specs for details...

Saving and loading from these should probably be optional... maybe it's the default option for `taskcluster.authenticate` I would be okay with that. This is sort of the thing people should want to do for most scripts.

jonasfj:
@jhford,

> but also happy to review or help write the credential storage stuff.

Please feel free to jump in and do the credential storage thing. trying both files and env vars would be nice... Obviously we need to read certificate to check if it has expired too...

jhford:
Let's file a bug for that as I won't have time to fix it too soon.

jhford:
https://bugzilla.mozilla.org/show_bug.cgi?id=1233805

PR #35, opened 2015-12-21
Read existing credentials if they're there

jhford:
this is just a work in progress for reading the credentials from the environment, then a configuration file as well as offering to save them when you do fetch them through the authenticate() method... let me know if you think this is horrible, I can't test it properly because I'm not able to VNC to my work desktop to click the button.  The code's a little messy :/

Please don't merge

### DIFF
--- a/taskcluster/utils.py
+++ b/taskcluster/utils.py
@@ -11,6 +11,8 @@ import sys
 
 MAX_RETRIES = 5
 
+CONFIG_FILE = os.path.expanduser('~/.config/taskcluster.json')
+
 log = logging.getLogger(__name__)
 
 try:
@@ -268,6 +270,45 @@ def isExpired(certificate):
   return expiry < int(time.time() * 1000 + 20 * 60 * 1000)
 
 
+def readCfgFile():
+  try:
+    with open(CONFIG_FILE) as cfgfile:
+      return json.load(cfgfile)['credentials']
+  except:
+    log.warn('Unable to read credentials file' + CONFIG_FILE)
+    raise
+
+
+def writeCfgFile(credentials):
+  data = {}
+  try:
+    try:
+      os.makedirs(os.path.dirname(CONFIG_FILE))
+    except:
+      pass
+    if os.path.exists(CONFIG_FILE):
+      with open(CONFIG_FILE) as cfgfile:
+        data = json.load(cfgfile)
+    data['credentials'] = credentials
+    with open(CONFIG_FILE, 'w') as cfgfile:
+      json.dump(data, cfgfile, indent=2)
+  except:
+    log.warn('Unable to write to config file ' + CONFIG_FILE)
+    raise
+
+
+def readCredentialsFromEnv():
+  c = {}
+  if 'TASKCLUSTER_CLIENT_ID' in os.environ and 'TASKCLUSTER_ACCESS_TOKEN' in os.environ:
+    c['clientId'] = os.environ['TASKCLUSTER_CLIENT_ID']
+    c['accessToken'] = os.environ['TASKCLUSTER_ACCESS_TOKEN']
+    if 'TASKCLUSTER_CERTIFICATE' in os.environ:
+        c['certificate'] = os.environ['TASKCLUSTER_CERTIFICATE']
+    return c
+  else:
+    return {}
+
+
 def authenticate(description=None):
   """
     Open a web-browser to login.taskcluster.net and listen on localhost for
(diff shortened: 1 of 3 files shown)


jhford:
Ok, I fixed the issues and was able to test this.  I'm thinking of using utils.authenticate() in the main client now that it supports reading the creds from saved places!

jhford:
@jonasfj if you like this, let me know!  I think the merge button will do the wrong thing regarding the 0.0.32 version release.

taskcluster/utils.py
```diff
@@ -11,6 +11,8 @@
 
 MAX_RETRIES = 5
 
+CONFIG_FILE = os.path.expanduser('~/.config/taskcluster.json')
```
jonasfj: Maybe just use: https://pypi.python.org/pypi/appdirs
It's a small dependency, from what I can see it's pure python... There's probably libs with better XDG compliance... but I doubt they are cross platform, I saw some solutions dependent glib...

Note: `os.path.expanduser('~/.config/taskcluster.json')` works great on linux and mac, but it is less nice on windows.

taskcluster/utils.py
```diff
@@ -268,6 +270,45 @@ def isExpired(certificate):
   return expiry < int(time.time() * 1000 + 20 * 60 * 1000)
 
 
+def readCfgFile():
```
jonasfj: How about allowing config file as parameter, and defaulting it to `CONFIG_FILE`...

Maybe rename this `readCredentialsFile()`... I think we can make it a convention across client libs to use JSON format in `taskcluster.json` file, under `user_data_dir`...

taskcluster/utils.py
```diff
@@ -276,6 +317,23 @@ def authenticate(description=None):
     The description will be shown on login.taskcluster.net, if not provided
     a default message with script path will be displayed.
   """
+
+  # Let's first offer to read credentials from a configuration file then
+  # the environment
```
jonasfj: This won't work... imagine a user with temp creds in file, and perma creds in env vars... :)

How about we just read from env vars, if we get something we return it... otherwise, we try file, and then we try webbrowser thing...

taskcluster/utils.py
```diff
@@ -268,6 +270,45 @@ def isExpired(certificate):
...
+    c['clientId'] = os.environ['TASKCLUSTER_CLIENT_ID']
+    c['accessToken'] = os.environ['TASKCLUSTER_ACCESS_TOKEN']
+    if 'TASKCLUSTER_CERTIFICATE' in os.environ:
+        c['certificate'] = os.environ['TASKCLUSTER_CERTIFICATE']
+    return c
+  else:
+    return {}
+
+
 def authenticate(description=None):
```
jonasfj: How about `configFile` as option defaulting to `CONFIG_FILE`, and have that `configFile = None` means we won't attempt to load from config file...

Option to load from env would also be nice, we can default it to `True`, as that is what most people will want.

taskcluster/utils.py
jonasfj: We can also have an option to save the credentials to disk... so we can skip the step where we ask the user...

taskcluster/utils.py
```diff
@@ -353,4 +411,14 @@ def do_GET(self):
...
     server.handle_request()
+
+  if sys.stdout.isatty():
+      print "Retreived credentials successfully.\n\nType 'yes' to save to " + CONFIG_FILE
+      a = raw_input().strip().lower()
+      if a == 'yes':
+        print "Saving configuration file"
+        writeCfgFile(creds[0])
+      else:
+        print "Not saving configuration file"
```
jonasfj: Okay, a crazy idea would be to use the webserver to do a confirm() in javascript to see if we should save the configuration file...  of course we need to run `server.handle_request()` until we got the credentials and the browser have told us if the credentials should be saved... it's an option.. maybe we can leave it for later...

Probably a bit overkill..  he he :)

jonasfj:
Generally, I like this a lot... !!! My comments are mostly just nits.. Maybe we should opinionated and not do any of them..

taskcluster/utils.py
```diff
@@ -276,6 +317,23 @@ def authenticate(description=None):
...
+  except:
+    log.warn('Exception thrown while trying to read credentials from file')
+
+  try:
+    existingCred.update(readCredentialsFromEnv())
+  except:
+    log.warn('Exception thrown while trying to read credentials from environment')
+
+  if 'clientId' in existingCred and 'accessToken' in existingCred:
+    return existingCred
```
jonasfj: Oh, I almost forget to mention this... **We must check `isExpired`** if the credentials have a certificate.

Otherwise, we can return credentials that won't work, which isn't much fun...

PR #40, opened 2016-02-15
bug 1247728 - Use sessions for requests

jhford:
@jonasfj what do you think about this?

### DIFF
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -41,6 +41,13 @@ _defaultConfig = config = {
 }
 
 
+def createSession(*args, **kwargs):
+    """ Create a new requests session.  This passes through all positional and
+    keyword arguments to the requests.Session() constructor
+    """
+    return requests.Session(*args, **kwargs)
+
+
 class BaseClient(object):
     """ Base Class for API Client Classes. Each individual Client class
     needs to set up its own methods for REST endpoints and Topic Exchange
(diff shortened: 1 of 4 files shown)


jonasfj:
This looks solid... One minor feature request is the ability to pass the session object via the constructor.

I see there is no way default session...

Also I see no need for resetSession... People can do my_client.session = ... 
If they really want to... But if they do this, it must be because they really know what they are doing.

I might be missing something but I see no use case for resetSession... You always want connection polling... Probably even a global one, I know I'm tempted to suggest making a single globale default session. That's what we do in node...

jhford:
I sort of want to make it so that someone can have a new session without having to import requests directly, but i don't feel strongly about including this method or not.

I'll add the ability to pass in a session.  Do we really want a global session?  I feel like when I make a new API client that the object I have should be distinct from all others.

jonasfj:
> I sort of want to make it so that someone can have a new session without having to import requests directly, but i don't feel strongly about including this method or not.

Agree... you can accept `None` and then fallback to creating a new session...
Or we could have a global session variable declared at module level.

taskcluster/client.py
```diff
@@ -69,6 +76,11 @@ def __init__(self, options=None):
             log.debug('credentials key scrubbed from logging output')
         log.debug(dict((k, v) for k, v in o.items() if k != 'credentials'))
 
+        if session:
+            self.session = options.session
```
catlee: should this be: ```self.session = session```?

jhford: yes.  that was a typo from addressing review feedback.  this would still work if you pass in the session as part of options, but thats not the desired situation.

PR #41, opened 2016-02-18
testAuthenticate

escapewindow:
@jhford - This PR:

* calls the tests via tox and coverage, instead of calling nosetests directly
 * this will allow for multiple python version tests in my py3 branch
* `python setup.py test [args]` will now call tox with the optionally specified args
* linting happens after tox, so we get test output before formatting errors.
 * (I was commenting out code and having to fix the formatting before I could see test output)
* we run `coverage html` after the tests, so we can see test coverage in htmlcov/
* ProductionTest has been replaced by TestTestAuthenticate, which does a get and a post.
* test_has_no_spaces now works no matter how the output is sorted

Green on my linux docker image.  I still get a segfault on the encrypt call on osx, but I don't think that blocks.


jhford:
Overall, looks good aside from ProductionTests.  The tox stuff seems to work well, thanks for adding that!  I would prefer that linting happen before unit tests, since the main reason I use the linter is to find bugs, but I understand the practical reasons for wanting it after unit tests.

I also noticed that if I do 'make clean' and then 'make test', I have to manually install both coverage and virtualenv... is that expected?

You've deleted all of the integration tests in ProductionTests though.  I'm happy to replace the mock auth server node stuff with testAuthenticate, but the ProductionTests are important.  They are integration tests to ensure that the client actually works.

The critical tests in there are:

* we can call methods which require no scopes with an unauthenticated client
* we can call methods which require authentication with valid permacreds
* we can call methods which require authentication with temporary credentials generated by python client
* we can call real methods with both position and keyword args
* we can use a signed url built with the python client

These are the most important of all the tests, so let's make sure that they aren't deleted.

setup.py
```diff
@@ -19,6 +21,8 @@
     'psutil==2.1.3',
     'hypothesis',
     'pgpy',
+    'tox==2.3.1',
+    'coverage==4.0.3',
```
armenzg: Should this be added as optional packages? (e.g. pip install taskcluster-client[tests])
Otherwise, users of this package would be forced to get these two packages installed.

jhford: these are the test_require, which we use devDep.py to install using pip... they are not installed for anyone who isn't running tests.

armenzg: Thanks for pointing it out! I had not expanded the diff to notice.

escapewindow:
* I can move the linter back as long as it doesn't exit before it runs the unit tests.
 * You can break out of it at that point if you need.
 * However, any errors might scroll off screen.  You will see the non-zero exit code though.
* The production tests don't work for me because I don't have the right scopes, and every time we add some the next set of tests break.  If I can turn them off without turning off the testAuthenticate tests, I'm find readding.  I might do a NO_TESTS_OVER_WIRE or NO_PRODUCTION_TESTS check, either var will skip em, so I can run the testAuthenticate tests without production, and jenkins still skips both.

jhford:
If the linter doesn't kill the build, we might as well not run it since no one will care about its output.  I'd prefer it to be run after the unit tests with errors being fatal than run at the beginning with non-fatal errors.

The solution to the production tests is to get you credentials with the right scope, not disable the tests.  Maybe I can generate a set of temporary scopes from my perma creds that has the scopes you need for this.  It's also ugly, but I could also run them for you if you need them, there just might be a 24hour turn around on the results.

escapewindow:
> I also noticed that if I do 'make clean' and then 'make test', I have to manually install both coverage and virtualenv... is that expected?

I just did
* clean clone
* git checkout testAuth
* make test
* make clean
* make test

I didn't have to install coverage or virtualenv manually.
Is it not installing into your venv?  It should install as part of `python devDep.py`.

escapewindow:
Re: linter, I meant not kill the test.sh run, but still exit non-zero.
That allows for:
* linter runs first
* linter still causes non-zero exit on failure
* but i can still run the tests after commenting out a line without proper line length / spacing.

I can keep it as is though.

If you can give me an id/creds (or give the aki or asasaki@m.c account those scopes) to run prod tests, that'll be very helpful. Whenever I asked for them in #taskcluster I was prodded to replace the tests with testAuthenticate tests.

escapewindow:
I've addressed your review comments.

* I squashed my previous commits so the commenting out / uncommenting type diffs shouldn't show.
* I don't have an issue with make clean; make test.  Are you still seeing this?
* the correct scopes to run the prod tests would be helpful, though I now have a way to skip them
* linter still runs after tests, though I can change that as long as it doesn't exit before the tests run.  it can change the final exit status instead.
 * I haven't changed this yet, but can.

I'm going to start working on my py3 and async branches.  Hopefully we can get this PR where we're both happy with it soon.

escapewindow:
@jhford -
* Dustin gave me a clientId with the right scopesets, so now I'm able to run production tests
* I did get an error with missing coverage when I ran test.sh without activating the venv, but it worked with either `make test` or after activating the venv.
* Happy to make changes to this PR if you have new comments.  I'll keep working on the py3 branch til then.

djmitche:
Regarding

> The critical tests in there are:
> 
> * we can call methods which require no scopes with an unauthenticated client
> * we can call methods which require authentication with valid permacreds
> * we can call methods which require authentication with temporary credentials generated by python client
> * we can call real methods with both position and keyword args
> * we can use a signed url built with the python client

I think we can divide these into two parts.  Some relate to URL construction, which can be validated locally, using something like https://github.com/djmitche/webmock or (for (signed) URL generation, by just validating the URL.

The remainder relate to correct construction of Hawk credentials, and those can all be tested against `testAuthenticate` and `testAuthenticateGet` with no need for production credentials.

I'll get to work on that now.

escapewindow:
@djmitche - I do have a TestTestAuthenticate class written, though I don't know if it covers the tests you want.  Feel free to use or discard as you see fit.  https://github.com/escapewindow/taskcluster-client.py/blob/d24b7d38a153cc1a3af4cf0164dcb07bb2d52de2/test/test_client.py#L600

djmitche:
@jhford please have a look at https://github.com/escapewindow/taskcluster-client.py/pull/1?

escapewindow:
@jhford any ETA on a review?

jhford:
Taking a look now.

jhford:
This looks good to me.  Thanks for getting rid of that node mess, i never liked having to do that!  (Fwiw, we used to also use the node installation from here to parse the apis.js file from the node library :-1: )

jhford:
why does having three consecutive zeros in a commit id amuse me so much?

djmitche:
Oh, I pushed the change to drop that to my branch, not aki's.  I'll push it to master.

PR #43, opened 2016-02-23
Support named temporary credentials

djmitche:
This adds support for creating named temporary credentials to the `createTemporaryCredentials` function.  I don't particularly like this function signature, but it doesn't leave any other options. I don't see the function documented anywhere, so I didn't update anything except the docstring (which inaccurately described the existing options)

### DIFF
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -529,14 +529,15 @@ def createApiClient(name, api):
     return type(name.encode('utf-8'), (BaseClient,), attributes)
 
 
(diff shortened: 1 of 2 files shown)


jhford:
Thanks for writing this support!  I'd like to merge this into master branch as a commit on the py3 work that @escapewindow is doing right now.  I've rebased the patch and put that into a PR against aki's repo.  If you have a second to make sure that I didn't screw up the merge, it'd be appreciated.

djmitche:
I didn't, but if you have the tests merged and they pass, then likely the rest is correct.

PR #45, opened 2016-03-03
generate code for taskcluster.sync.*

escapewindow:
Hey @jhford,

This is a pretty big PR, so I don't expect a super fast turnaround.
However, the sooner you can alert me to big problems, the less I'll have to rework my upcoming async work.  If you can let me know your ETA/progress, that would be great, thank you!

### High Level Overview

We now generate the api code in `taskcluster.sync` via `make gencode`.  These files are committed to git, the same as apis.json.  That allows us to inspect the files and see any changes easily.

`taskcluster.client` now imports all the api objects from `taskcluster.sync`.  The existing tests have been modified to allow for this.  New generated code tests check for the existence of all of the appropriate objects and files, and make sure the calls to `_makeHttpRequest` and `_makeTopicExchange` are sane.  Each generated method is limited to 2-4 lines of python code.  The generated code files have 100% test coverage and pass flake8.

Because the code and apis.json are tied more closely now, we'll need to run `make update` more often to use the most recent api.  Eventually I'd like to see an automated update / test / dot release whenever the api definitions change.

I named it `taskcluster.sync` rather than `taskcluster.generated` because I'm planning on creating `taskcluster.async` soon.  The `taskcluster.Queue` and `taskcluster.client.Queue` etc. objects will be sync or async depending on your version of python (due to the imports + `__all__` hacking in `taskcluster/__init__.py` and `taskcluster/client.py`), but if you want to explicitly use the sync or async objects, you can use `taskcluster.sync.*` or `taskcluster.async.*`.

### Nitty Gritty Details

* `Makefile` changes:
 * I split out `make test` into `make nosetests` and `make lint`.  Now you can test linting or nosetests in any order, although `make test` runs nosetests first.
 * Now that the Makefile runs nosetests, coverage, and linting, there's no real reason to keep `test.sh` around.  I nuked it.
 * I added a `gencode` target, and made it part of `make update`.

* `devDep.py` and `setup.py` changes:
 * I added a `gencode_requires` list for jinja2.

* `genCode.Py` - new file that generates the code and test files.
 * I named it `genCode.py` to match `genDocs.py`
 * Let me know if you need more comments/docs on this

* `baseclient.py`
 * I split out `BaseClient` into `taskcluster.baseclient`, because:
  * the generated code needs to import `BaseClient`
  * `taskcluster.client` imports the generated code
  * if `BaseClient` is in `taskcluster.client`, that creates a circular import dependency chain that barfs

* `client.py`
 * There is now a `RuntimeClient`.
 * Essentially, if I used `_makeApiCall` in the generated code, that required me to pass around a full copy of the api definition.  The Releng team also prefers having explicit arguments in each method, rather than `*args` and `**kwargs` everywhere.  So `BaseClient` now has the base logic, and `RuntimeClient` can take the api definition and translate it to `BaseClient`.
 * Nothing uses `RuntimeClient` now, but I remember you had a requirement to allow for `createApiClient()` calls at runtime, and keeping it around allows for that.
 * We import all api classes from `taskcluster.sync`, and fall through to `createApiClient()` if it doesn't exist.  However, if any don't exist, the `test_generated.py` tests will fail.

* `taskcluster/sync/[A-Z]*.py`
 * These are all generated, so feel free to use whatever level of review scrutiny you think is appropriate for these.
 * The `routingKeys` only copies a subset of the routingKeys.  This is largely for file size and line length considerations.  We only use a subset of the keys, so I included the ones we use.  When we use more, we can add them to the `code.template` and `test/base/__init__.py` whitelists.

* `templates/*.template`
 * `code.template` is for the `taskcluster.sync` files other than `__init__.py`, and
 * `test.template` is for the generated test files other than `test_generated.py`.

* `test/base/__init__.py`
 * I added support for the generated code tests here.
 * `FakeGenerated` is intended for multiple inheritance, e.g. `class FakeAuth(FakeGenerated, Auth):`
  * It replaces `_makeHttpRequest` and `_makeTopicExchange` with `MagicMock()` objects
  * `buildSignedUrl` returns `url + 'SIGNED'`
  * Essentially, this allows for testing code paths and args without testing the underlying logic.
 * `GeneratedTC` has generic methods to test topic exchanges and signed/unsigned functions.  It also tests `self.urls` and `self.routingKeys` against apis.json.
 * This file could have shared more code with `genCode.py`, but I wanted to avoid the case where we use a broken function to test itself.

* `test/test[A-Z]*.py`
 * These are the generated test files, one per generated code file.

* `test/test_client.py`
 * This required a few changes to account for the `RuntimeClient` / `BaseClient` split.
 * If you open `htmlcov/index.html` after running `make test`, you'll see all the generated files have 100% coverage.

* `test/test_generated.py`
 * This uses `apis.json` to make sure all apis have generated classes in `taskcluster.sync`, all apis have files in `taskcluster/sync/`, and no extra .py files are in `taskcluster/sync/`.


escapewindow:
@djmitche, @jonasfj, I'd love to hear your thoughts as well, if you have the time/inclination.  Or if there's someone I missed that should take a look.

taskcluster/sync/Auth.py
```diff
@@ -0,0 +1,486 @@
...
+    services. These API end-points are of interest if you wish to:
+      * Authenticate request signed with TaskCluster credentials,
+      * Manage clients and roles,
+      * Inspect or audit clients and roles,
+      * Gain access to various services guarded by this API.
+    ### Clients
+    The authentication service manages _clients_, at a high-level each client
+    consists of a `clientId`, an `accessToken`, scopes, and some metadata.
+    The `clientId` and `accessToken` can be used for authentication when
+    calling TaskCluster APIs.
```
djmitche: Is this missing double-newlines to delimit paragraphs?

djmitche:
From a brief look, this is very nice!

taskcluster/sync/Auth.py
escapewindow: Good call.  Committed e340e7a77 to fix.

escapewindow:
Thanks for the feedback @djmitche!

taskcluster/sync/Auth.py
```diff
@@ -0,0 +1,511 @@
...
+        Get a list of all clients.  With `prefix`, only clients for which
+        it is a prefix of the clientId are returned.
+
+        This method takes no arguments.
+        '''
+        url = self.urls['listClients'].format(
+            baseUrl=self.options['baseUrl'],
+        )
+        if signUrl:
+            url = self.buildSignedUrl(url)
```
jonasfj: What exactly are you trying to do here?

From looking at `_makeHttpRequest` it seems that if credentials are provided it will sign the HTTP request using hawk authorization header as expected.

The purpose of url signing is for the client to give you a signed url... Not call the API using a signed url.
If downloading large files or granting someone else access signed urls are very nice.

We could extend the client and reference format to support large binary files, but it's a lot of work to get right. curl is already good at it.

taskcluster/baseclient.py
```diff
@@ -0,0 +1,324 @@
...
+            payload = utils.dumpJson(payload)
+
+        # Do a loop of retries
+        retry = -1  # we plus first in the loop, and attempt 1 is retry 0
+        retries = self.options['maxRetries']
+        while retry < retries:
+            retry += 1
+            # if this isn't the first retry then we sleep
+            if retry > 0:
+                snooze = float(retry * retry) / 10.0
```
jonasfj: My math could be wrong but I think this is only polynomial...

In node client I have used (inspired by aws):
`delay = max(pow(2, retry) *delayFactor * random([1 - randomizationFactor; 1 + randomizationFactor]), maxDelay)`

with `delayFactor = 100`, `randomizationFactor = 0.25` and `maxDelay = 30000`.

Easily implemented as:
https://github.com/taskcluster/go-got/blob/031f55c680095b926505a79d0e2fa0789e637970/backoff.go#L24-L29

`delay = max(pow(2, retry) * delayFactor * (randomizationFactor * (random() * 2 - 1) + 1), maxDelay)`
Where `random()` is between 0 and 1.

taskcluster/sync/Auth.py
escapewindow: IIRC, I asked in #taskcluster which methods would need to support signed urls, and you told me it's optional for gets, and the others don't support it.  Because of that, I added the capability to sign the url for all gets.

I don't have any other criteria for which make sense to sign or not.  Are there ways for me to programmatically decide which should support signed urls?

taskcluster/baseclient.py
```diff
@@ -0,0 +1,324 @@
...
+            else:
+                value = '#' if key.get('multipleWords') else '*'
+                log.debug('Did not find %s in input params, using %s', key['name'], value)
+
+            routingKeyParts.append(value)
+
+        data['routingKeyPattern'] = '.'.join([utils.toStr(x) for x in routingKeyParts])
+        return data
+
+    def buildSignedUrl(self, requestUrl, expiration=None):
```
jonasfj: We missing `buildUrl` and this doesn't really do what we want...
See: https://github.com/taskcluster/taskcluster-client#construct-urls

It's probably easiest do this with `*args` and apply format...

`url = queue.buildUrl('task', taskId, expiration)`
and 
`url = queue.buildSignedUrl('task', taskId, expiration)`

To build urls for the `task` method.

taskcluster/baseclient.py
escapewindow: That's verbatim from https://github.com/taskcluster/taskcluster-client.py/blob/master/taskcluster/client.py#L364 .

Once I go through `git blame` and skip past my flake8 whitespace changes, it looks like this is from your commit: https://github.com/taskcluster/taskcluster-client.py/commit/0533cbe8de5623151694e151fbae024049ac2351#diff-cef37b865d30b6dc7000a5e7c28d3fd6R330

taskcluster/baseclient.py
jonasfj: Also this is complicated by optional querystring parameters... 
Those are relatively new so not well documented in the docs anywhere..

escapewindow: RuntimeClient has buildUrl if you need it.  BaseClient is its base, and expects a url to be passed to it.  The generated code has all the urls.

jonasfj:
Generally, this looks good.

> The `taskcluster.Queue` and `taskcluster.client.Queue` etc. objects will be sync or async depending on your version of python

As neat as it sounds, and yes it is really tempting... Maybe consider if perhaps this is too smart.
It's effectively creating a incompatibility between py2 and py3...

> The routingKeys only copies a subset of the routingKeys.

1. This is not used, because pulse doesn't really have decent python client
2. You hopefully copy over all routing keys, but not necessarily all their properties, like summary (that's fine)
3. References are used for both docs and clients, entries like "description" is more important for docs.

----
Regarding query-string support. The reference format specify what query string parameters are allowed.
These are always optional, they are always strings.

For node usage see:
https://github.com/taskcluster/taskcluster-client#providing-options

Basically, something like: `def listTaskGroup(self, taskGroupId, options = {}):`

So that calling it is `queue.listTaskGroup(taskGroupId)` or `queue.listTaskGroup(taskGroupId, {"continuationToken": ..., "limit": ...})`. Query string options are always optional, they are often used for listing...

escapewindow:
>> The taskcluster.Queue and taskcluster.client.Queue etc. objects will be sync or async depending on your version of python

> As neat as it sounds, and yes it is really tempting... Maybe consider if perhaps this is too smart.
It's effectively creating a incompatibility between py2 and py3...

I think the classes should have the same arguments.  If not, I will definitely rethink this.
I'm not terribly concerned because people can explicitly import and use `taskcluster.sync.*` if they want to use the synchronous py2+py3 supported classes.  I'll give this more thought as the async code gets fleshed out.

>> The routingKeys only copies a subset of the routingKeys.
> This is not used, because pulse doesn't really have decent python client

I'm not sure what this means.  Does this mean the python client doesn't need to support topic exchanges?

> You hopefully copy over all routing keys, but not necessarily all their properties, like summary (that's fine)

Hm, that's easy enough to add, but I'm not sure what would that give us.
I'm referring to https://github.com/taskcluster/taskcluster-client.py/pull/45/files#diff-fae555850972db367d99fb4ef03f3771R22 , which is used when we call `self._makeTopicExchange` https://github.com/taskcluster/taskcluster-client.py/pull/45/files#diff-fae555850972db367d99fb4ef03f3771R95 .  What is the benefit of having 'summary' as a key with an empty value?

If you're worried about people using that routingKeys definition instead of apis.json, maybe I should insert a comment in the generated code saying that this is only a subset of the routingKeys definition, and reference the api definition for the full set.  Would that work?

> References are used for both docs and clients, entries like "description" is more important for docs.

I'm only copying over these three keys and their values: https://github.com/taskcluster/taskcluster-client.py/pull/45/files#diff-e3888f431c8c5d19f0eb42906072c5fbR29
Am I missing anything important?  That's all the client uses currently.

I'll take a look at adding options support.  Thanks for the feedback and link!

taskcluster/sync/Auth.py
jonasfj: > IIRC, I asked in #taskcluster which methods would need to support signed urls....

Oh, yeah I remember that conversation I didn't really understand what you were doing. Sorry, we should have clarified that, before the discussion got side tracked on something else...


----
Let me clarify, the ability to make signed urls is not an authentication mechanism the client library should employ when making requests on behalf of a client. Hawk bewit or signed-urls, whatever you call them are less secure.

It is a feature similar to the ability to generate temporary credentials. The client library should expose a method that allows the user to generate a signed url for an API end-point (not calling with a signed url). The user can then get the signedUrl as a string, and pass it off to someone else.

If downloading a 2GB artifact maybe it's better to use curl with the correct settings, than trying to streaming things through a python library (particularly when the python library tries to parse the body as json).

taskcluster/baseclient.py
jonasfj: he he, yeah we did it wrong in initial version of node client too :)

taskcluster/sync/Auth.py
escapewindow: Ok.  I can tear out the signUrl options pretty easily, and the buildSignedUrl method is there already.  Thanks for the explanation!

taskcluster/baseclient.py
jonasfj: From: http://references.taskcluster.net/queue/v1/api.json

```js
    {
      "type": "function",
      "method": "get",
      "route": "/task-group/<taskGroupId>/list",
      "query": [
        "continuationToken",
        "limit"
      ],
      "args": [
        "taskGroupId"
      ],
      "name": "listTaskGroup",
      "stability": "stable",
      "title": "List Task Group",
      "description": ...
```

See the `query` array... It's still not on all reference documents. But just assume it to be empty array when not present.

jonasfj:
> I'm not sure what this means. Does this mean the python client doesn't need to support topic exchanges?

Just say, let's not worry too much about it... because nothing uses these methods yet.

>> You hopefully copy over all routing keys, but not necessarily all their properties, like summary (that's fine)
> What is the benefit of having 'summary' as a key with an empty value?

Nothing, `summary` is a property of a routing key entry. Hopefully you copy over all the routing key entries (I don't care about the properties they have for documentation, like summary). From what I can see you don't filter them...

From a quick look at: https://github.com/taskcluster/taskcluster-client.py/pull/45/files#diff-ac144115cad0259dba88b6251ee419d0R125
It you seems you got it right :)
So nvm...

taskcluster/baseclient.py
escapewindow: Ok, thanks for clarifying!  I'll try to add a buildUrl, though I still want to avoid passing around the entire api definition in BaseClient.
The non-exchange objects already have a self.urls with string formatting; I think that, plus allowing for query support, may be sufficient.

escapewindow:
I think I've addressed the issues in my new code.
* `signUrl` support is removed from the generated code, even though it was off by default
* there is now support for `buildUrl` in `baseclient`.
* I fixed a bug where I was passing the full url to _makeHttpRequest when it was expecting a route.

These are existing bugs that I haven't touched [yet]:
* query strings and options don't seem to be supported in the current taskcluster-client.py, though I might be wrong
* the exponential backoff math on retries is an existing bug

Both should be fixed, but I think they shouldn't block this PR if they're bugs that existed before I started this branch.

escapewindow:
> These are existing bugs that I haven't touched [yet]:
>    query strings and options don't seem to be supported in the current taskcluster-client.py, though I might be wrong
>    the exponential backoff math on retries is an existing bug

I have addressed these with the latest commits.

genCode.py
```diff
@@ -0,0 +1,169 @@
...
+import re
+import six
+import textwrap
+
+from jinja2 import Environment, FileSystemLoader
+
+GENERATED_STRING = "# This file is generated!  Do not edit!"
+# ROUTING_KEY_WHITELIST is also defined in test/base/__init__.py
+# if you change it, please update it there as well.
+ROUTING_KEY_WHITELIST = ("name", "multipleWords", "constant")
```
jhford: I have a strong aversion to having the same data defined in two different places that need to have a human keep them in sync.  Can you instead put this somewhere that both locations can reference?

genCode.py
```diff
@@ -0,0 +1,169 @@
...
+def createInitPy(pydir, modules):
+    initpy = os.path.join(pydir, '__init__.py')
+    print(initpy)
+    with open(initpy, 'w') as fh:
+        print("#!/usr/bin/env python", file=fh)
+        print(GENERATED_STRING, file=fh)
+        for module in modules:
+            print("from taskcluster.sync.{module} import {module}".format(module=module),
+                  file=fh)
+        print("__all__ = [", file=fh)
```
jhford: Do we really need `__all__` in this case?  `__all__` only tells which module-level definitions from this module should be imported when you do `from x import *` right?  In this case, we're just exposing all things that are in module-level scope.  In the old __init__ file, we were limiting what was being exposed by using __all__

taskcluster/client.py
```diff
@@ -603,5 +295,12 @@ def createTemporaryCredentials(clientId, accessToken, start, expiry, scopes, nam
...
 # This has to be done after the Client class is declared
 for key, value in API_CONFIG.items():
-    globals()[key] = createApiClient(key, value)
+    # import buildtime generated code if it exists; otherwise use runtime generated code
+    try:
+        module = __import__('taskcluster.sync.{key}'.format(key=key), globals(),
+                            locals(), [key], 0)
+        globals()[key] = getattr(module, key)
+    except (ImportError, AttributeError):
+        print("{} not in taskcluster.sync!!!".format(key))
```
jhford: This is planning for a case that I don't think is likely.  If a service is in the Apis.json file, it will be generated in taskcluster.sync.  I'd rather just export the `createApiClient` method and force the export to use the taskcluster.sync methods.

taskcluster/client.py
```diff
@@ -2,231 +2,24 @@
...
-        u = urllib.parse.urlparse(requestUrl)
-        return urllib.parse.urlunparse((
-            u.scheme,
-            u.netloc,
-            u.path,
-            u.params,
-            u.query + 'bewit=%s' % bewit,
-            u.fragment,
-        ))
+class RuntimeClient(BaseClient):
```
jhford: I think it's sort of weird for this class to exist in the client.py file, when it's not the default used in client.py to export.  Can you maybe move this and `createApiClient` into a new file, maybe taskcluster/runtimeclient.py then import createApiClient and add it to this module's globals and __all__?

genCode.py
jhford: If the purpose of this is to remove a couple specific keys, I think a blacklist would be better.

genCode.py
```diff
@@ -0,0 +1,169 @@
...
+                        value = routingKey[key]
+                        if isinstance(value, (six.binary_type, six.text_type)):
+                            value = "'{}'".format(value)
+                        parts.append("            '%s': %s," % (key, value))
+                parts.append("        },")
+            parts.append("    ],")
+    return '\n'.join(parts)
+
+
+def methodArgumentString(entry):
```
jhford: This comment is applicable to methodArgumentString, argumentNames, argumentString.  These methods would be really great if they included a brief description of why they exist.  In this case something like "This function is used to generate a string version of an argument list for a stub method in an API class."

genCode.py
```diff
@@ -0,0 +1,169 @@
...
+    """Returns an argument string for the given function
+    """
+    parts = ['self']
+    parts.extend(argumentNames(entry))
+    if entry.get('query'):
+        parts.append('options=None')
```
jhford: Why isn't this part of `argumentNames`?  because of the default value? Do we really need this to be two separate functions?

genCode.py
```diff
@@ -0,0 +1,169 @@
...
+    parts = []
+    if 'args' in entry:
+        parts.extend(entry['args'])
+    if 'input' in entry:
+        inputName = 'payload'
+        parts.append(inputName)
+    return parts
+
+
+def argumentString(entry):
```
jhford: Why do we need this function at all?

What about:

`def generateArguments(entry, includeSelf=False):`

?

genCode.py
```diff
@@ -0,0 +1,169 @@
...
+        methodArgumentString=methodArgumentString,
+        argumentString=argumentString,
+        createRoutes=createRoutes,
+        createRoutingKeys=createRoutingKeys,
+        generatedString=GENERATED_STRING,
+        referenceUrl=url,
+    )
+
+
+def docstringify(s, level=4):
```
jhford: Can we use a more descriptive name, which doesn't use the ..ify suffix?  `typesetDocstring` would work?

templates/code.template
```diff
@@ -0,0 +1,100 @@
...
+
+class {{serviceName}}(baseclient.BaseClient):
+    '''
+    {{api['title']}}
+    {{api['description'] | docstring}}
+    '''
+    version = {{api['version']}}
+    referenceUrl = '{{referenceUrl}}'
+{%- if createRoutes(api) %}
+    routes = {
```
jhford: Curious why these are class variables instead of being baked into the stub methods as literals.

test/base/__init__.py
```diff
@@ -100,3 +113,114 @@ def forNode(self):
             'expires': self.expires,
             'scopes': self.scopes
         }
+
+
+class FakeGenerated(object):
```
jhford: Multi inheritance doesn't feel like the right approach here.  Could we instead have a method like
```python
def createFakeApi(cls, *args, **kwargs) {
  instance = cls(*args, **kwargs)
  instance._makeHttpRequest = mock.MagicMock()
  instance._makeTopicExchange = mock.MagicMock()
}
```
?

jhford:
Overall, this looks good.  Once the review issues are all addressed, I'd be happy to merge this.  A couple things which don't have specific line numbers to reference are:

1. I remember a discussion about doing something like py2 defaulting to sync and py3.5+ defaulting to async... I don't know if that's true, but I don't think this is something we should do.

2. There are no tests which ensure that createApiClient created classes work.  What would be really cool is that the over-wire unit tests (i.e. integration tests) would all be run for both the build and run time generated api classes.

Sorry to not have gotten this back to you last week, I was most of thursday and friday.  Let me know if you have any questions or wish to discuss anything.  I'm in PST for this week, so that should be easier to do.

genCode.py
escapewindow: I'm able to get around the duplicate, as long as you don't mind `sys.path` munging.

genCode.py
escapewindow: You're right, it's not really needed.
However, at the very least, it's suppressing lint warnings.  Given a choice between

* `# NOQA` or other lint skipping methods, or
* keeping `__all__`,

which do you prefer?  I prefer the latter.

djmitche:
Just to chime in, I'm also opposed to different defaults for py2 and py3.  "explicit is better than implicit" :)

escapewindow:
> Just to chime in, I'm also opposed to different defaults for py2 and py3. "explicit is better than implicit" :)

I agree explicit > implicit.
However, that would lead me to say, we should all use taskcluster.sync.* or taskcluster.async.* rather than importing taskcluster.sync.* into the taskcluster.client and taskcluster namespaces.

genCode.py
escapewindow: This is for `test.template`: https://github.com/taskcluster/taskcluster-client.py/pull/45/files?diff=unified#diff-61eaabe68b94e8221213d0fa4559e5a8R39

The `methodArgumentString` also has `options=None`, so includeSelf is a misnomer.  I'll try to figure out the right name and add docstrings.

djmitche:
> However, that would lead me to say, we should all use taskcluster.sync.* or taskcluster.async.* rather than importing taskcluster.sync.* into the taskcluster.client and taskcluster namespaces.

That sounds OK to me.

escapewindow:
So far I've gotten an r- for removing the taskcluster.* namespace objects, because it'll break existing code.

Catlee was suggesting I split this into two different packages, generated from the same repo  (`make gencode-async`?).  I think this would be much more difficult to do, but would allow for both explicitly choosing to use async while [I *think*] making it fairly transparent to drop one or the other into your code.

jhford:
I think that having the default of `from taskcluster import Index` be the sync class that doesn't change the api for existing code is the right thing to do.  I guess we could have a second module for the async stuff, but that seems like a really heavy hammer for something that's really quite simple:

* if you want synchronus client objects, `import taskcluster`
* if you want async client objects, `import taskcluster.async`

Is there a reason to make it more complicated than that?

Another option, in the style of promise-wrappers for JS would be to have a method name suffix for the async version and have one client object able to do sync and async.  I don't really think this is the way to go, but I just wanted to put it out there as an option.  The only confusion is if we had an endpoint which ended with `Async`.  In that case, the sync call would be `badNameAsync` and the async version would be `badNameAsyncAsync`.  That's one reason I'm not keen on this idea.  An example would be:

* `index.listNamespaces()`
* `index.listNamespacesAsync()`

escapewindow:
Hm, or just an `async=True` in the `__init__`.
I agree that this is not a huge issue.  It does seem like a structure of taskcluster.* vs taskcluster.async.* makes one a second class citizen, but as long as that's not the case in practice, perhaps I should just let it go.

jhford:
I like that idea as well.  My motivation here is that someone who wants the most simple option possible to run a taskcluster api method is able to do it without too many roadblocks and not breaking api where needed.  I'd rather use the async methods myself, but i think sync being the default is a good choice.

genCode.py
escapewindow: I wasn't sure how to answer that, so I combined them: https://github.com/escapewindow/taskcluster-client.py/commit/eb0696f77723d568fe40c9b7a736a15ecf41335f

The answer is `test.template` needs the `argumentString` single quoted without self or the `kwargs` (only `options`, atm).  It makes the function a bit more complicated, but I think it's ok.  Let me know if that looks ok.

templates/code.template
escapewindow: I wanted to be able to use them in `baseclient` and tests.

genCode.py
escapewindow: As mentioned above, https://github.com/escapewindow/taskcluster-client.py/commit/eb0696f77723d568fe40c9b7a736a15ecf41335f .  `methodArgs=False` was my choice there.

taskcluster/client.py
escapewindow: Easy enough, and I agree.  Done.

taskcluster/client.py
escapewindow: +1, done.  `test_generated.py` already tests for missing classes, so this is unneeded.

genCode.py
jhford: i don't know if that's a *great* place for it to go, but it does beat having a person needing to keep them in sync or using sys.path stuff.

escapewindow: The blacklist is looking good.  We're not using `required`, but other than longer files, it's harmless to include.  https://github.com/escapewindow/taskcluster-client.py/commit/8f8970277a91d451b53863fad52fa376301e2a85

(I'm going to be merging all of this into the codegen branch, and this PR, once I've dealt with all the review comments.)

test/base/__init__.py
escapewindow: Sure. https://github.com/escapewindow/taskcluster-client.py/commit/b5075a822eb16b8085c4293898407b11ac74d896

I think an earlier revision of `FakeGenerated` also overrode several other methods with non-mock code, until I rewrote things and didn't need them anymore.  Easy enough to replace with a function now.

genCode.py
escapewindow: Done.

escapewindow:
Current status:
* merged my current work into the `codegen` branch
* I have not removed the `__all__` from `test/sync/__init__.py` because flake8.  If you feel strongly about it, I can replace it with `# NOQA` comments.
* I have not yet added `createApiClient` tests, but am planning on doing so.  I think a number of the existing tests use some of that framework already, but I can look into the integration tests spanning both.

genCode.py
jhford: I like the new version.

jhford:
great!  it's looking good.  My only concern with `__all__` was that it was there for no reason.  not turning off the linter there is worth having it there.  Good to hear that you're going to have createApiClient tests.  I know that we use it in the unit tests, but I want to make sure that we make sure that actual real api methods can be called with it as well.

This PR is getting quite full and has at least two or three sets of feedback in it.  Would you be OK with us closing this PR and opening a new one for the remaining work here?

escapewindow:
Sure, that works for me. I'll refer to this PR in the next one.

PR #46, opened 2016-03-08
generate code for taskcluster.sync.*

escapewindow:
This is a continuation of #45.


escapewindow:
@jhford how does https://github.com/escapewindow/taskcluster-client.py/commit/92d80062a9cd841be045e4a1990186dcace61802 look for dual integration tests?

I'm leaning towards splitting that out into a test_integration.py, but I don't feel strongly about it.

escapewindow:
I thought I should mention, since I would wonder if I were in your position:

I wanted to have `TestRuntimeAuthentication` inherit `TestAuthentication`, only replacing the `taskcluster.Auth` class with a `createApiClient` class.  That would be a lot neater and simpler.  However, when I did that, nosetests didn't run any tests for `TestRuntimeAuthentication`.

Maybe there's a trick I don't know, but this was the way I got tests to run for both `TestRuntimeAuthentication` and `TestBuildtimeAuthentication`.

jhford:
Looks good!  It might be possible to generate the classes programatically using the type function, (see createApiClient as example), but I don't think that's blocking.  I'd like to take one more overall look at this a little later, but I don't expect there to be anything major.

Thanks for working on this!

escapewindow:
@jhford, have you had a chance to take another look yet?

jhford:
Yep!  Looks good.  I'm going to merge this but hold off on releasing a new major version to send an email suggesting that users of the client library try it out and let us know if there are issues for them.

PR #47, opened 2016-03-11
createTemporaryCredentials takes datetime objects

djmitche:
Also remove some unused code


jhford:
this will land as part of other work as e52ccd2

PR #53, opened 2016-05-04
pgpy 0.4.0 !

escapewindow:
While adding py3 support to taskcluster-client.py, I ran into https://github.com/SecurityInnovation/PGPy/issues/154 , which munged messages containing unicode characters before encrypting.  My workaround involved explicitly converting the message to a bytearray before encrypting and after decrypting:  https://github.com/taskcluster/taskcluster-client.py/commit/63ccdc3a88af142862374c2b3641973e2024f937#diff-ba298f6131fc96148efc57c00e72f654R252

https://github.com/SecurityInnovation/PGPy/issues/154 was fixed in late February, but there was no new PGPy release until two weeks ago, so there was no package on pypi containing the fix.

Now that PGPy 0.4.0 is released with this fix, we can depend on it and remove the workaround.  This is mainly a cleanliness fix; this PR is not blocking anything.

The reason I'm specifying `pgpy>=0.4.0` is that the pypi version is actually `0.4.0.post1`.  If we want to pin it to a single version, I can specify `pgpy==0.4.0.post1`.  We're not currently pinning pgpy.


jhford:
This was addressed as part of the v1.0.0 release

PR #56, opened 2016-09-16
Suport hawk header directly in credentials.

La0:
Hello TC maintainers,

I'm working on a [Mozilla project](https://github.com/mozilla-releng/services/) using TaskCluster auth from a web app in 2 parts:
- a javascript frontend gets and stores the TC credentials. It can then build some hawk authentication header
- a python backend uses TC authenticateHawk to authorize the requests from the frontend

The project needs to use more services from TaskCluster (secrets, queues, ...) and in order to avoid transmitting the full TC credentials to the backend, we'd like to send directly the hawk header for a request to the TC python client.

So this Pull request adds an extra credential option, permitting to use an existing hawk header.

Thanks for your review.

### DIFF
--- a/taskcluster/baseclient.py
+++ b/taskcluster/baseclient.py
@@ -30,7 +30,7 @@ _defaultConfig = config = {
         'clientId': os.environ.get('TASKCLUSTER_CLIENT_ID'),
         'accessToken': os.environ.get('TASKCLUSTER_ACCESS_TOKEN'),
         'certificate': os.environ.get('TASKCLUSTER_CERTIFICATE'),
-        'hawkHeader' : None,
+        'hawkHeader': None,
     },
     'maxRetries': 5,
     'signedUrlExpiration': 15 * 60,
--- a/taskcluster/baseclient.py
+++ b/taskcluster/baseclient.py
@@ -30,6 +30,7 @@ _defaultConfig = config = {
         'clientId': os.environ.get('TASKCLUSTER_CLIENT_ID'),
         'accessToken': os.environ.get('TASKCLUSTER_ACCESS_TOKEN'),
         'certificate': os.environ.get('TASKCLUSTER_CERTIFICATE'),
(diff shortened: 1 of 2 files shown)


La0:
Now the PR passes all checks. Could you please review it ?

Thanks !

taskcluster/baseclient.py
```diff
@@ -256,6 +257,15 @@ def _hasCredentials(self):
...
         )
 
+    def _hasHawkHeader(self):
+        """ Return True, if hawk header is already given """
+        cred = self.options.get('credentials')
+        return (
+            cred and
+            'hawkHeader' in cred and
+            cred['hawkHeader']
+        )
```
jhford: please rewrite this as a more clear structure.  Having two checks (in and truthy) are redundant in Python since we have the set-membership (`in`) feature.

walac: yeah, this can be simplified to:
```python
return cred and cred.get('hawHeader', False)
```

La0: I just pushed the simplified version.

La0:
Hello @jhford,

Can you please review again my PR ? I updated it with a new commit.

Thanks !

jhford:
I think in this case, you could use `buildUrl` to build a URL and then generate whichever extra headers you want.  The `buildUrl` method is not documented in the README.md file (but should be).  It's located here:

https://github.com/taskcluster/taskcluster-client.py/blob/master/taskcluster/client.py#L154

It should be something like

```python
import taskcluster
index = taskcluster.Index()
url = index.buildUrl('listNamespaces', 'namespacetolist')
```

This will assign url to the URL to hit.  Then you could use the `utils.makeHttpRequest` function if you'd like things like backoffs, etc.

```python
import taskcluster.utils
# assuming url figuring from above
result = taskcluster.utils.makeHttpRequest('get', url, None, {'headers': 'as-dict'})
```

PR #61, opened 2017-01-30
Fix missing async in package

jhford:
This will package the async module for taskcluster.  We cannot use conditionals in `setup.py` to only package `taskcluster.async` on python <= 2 because the output of `setup.py` is used to decide which files to include in the tarball which gets uploaded to pypi.  Instead, we need to make sure that we ignore the async libraries while testing, but explicitly *include* them in the packages kwarg to `setup()` from distutils, even in python 2.7.

Note: this will cause a bunch of ugliness in the install log on python 2.7, but that can safely be ignored.  It is addressed with https://github.com/pypa/pip/commit/09a74c60626b797ebfab951ec59f7b35e262ba2e.  These are simply byte compiling errors, which pip will ignore since version 6.0 if the installation otherwise succeeds .

Other changes:

* use tox-travis to split out the languages a little nicer.
* fixes to allow the build system to work nicely on OSX.  I'm using the python.org installers to get this to work.
* open the files which we're writing out in binary mode and explicitly encode the strings into bytes on both python versions.
* sort the function info dictionary to minimize future diffs.


jhford:
This should fix https://github.com/mozilla-releng/beetmoverscript/issues/38 and https://github.com/mozilla-releng/scriptworker/issues/70 when it lands and is released.

jonasfj:
@jhford, so now it'll be `import taskcluster` or `import taskcluster.async` depending on what client I want, right?

jhford:
yep!

And for added fun, you could actually `import taskcluster.auth` if you *really* only want the auth class ever.  That's more a side-effect of how the module loads itself than a conscious choice though ;)

escapewindow:
Thanks!

PR #64, opened 2017-02-22
Add new auth azure functions. This is a breaking change.

imbstack:
For context on why this is breaking, you can start reading from https://github.com/taskcluster/taskcluster-auth/pull/91#pullrequestreview-20124716

Other than that, this is just a change made by running `make update`. I'm happy to do the `release.sh` stuff once this is in master. Do I need creds for that?


jhford:
I'll do the release.sh stuff this time to double check it, since this is the first release since I rewrote it.

jhford:
Version 1.2.0 released!

PR #66, opened 2017-03-29
Await asyncio.sleep() calls

rail:
`asyncio.sleep()` returns immediately, so we should `await` on it before processing.


rail:
@jhford can you look at this?

jhford:
Looks good to me, thanks for the patch!

PR #69, opened 2017-04-21
README: Remove leftover mention of PyHawk

### LINKED ISSUE #54: Readme mentions bug in the no longer used PyHawk library
> There is a bug in the PyHawk library (as of 0.1.3) which breaks bewit
generation for URLs that do not have a query string. 

From: https://github.com/taskcluster/taskcluster-client.py/blame/c155902351b0c695c58516ebd83d769185c7f817/README.md#L79-L80

This can be removed, since taskcluster-client.py now uses mohawk instead of PyHawk.

gregarndt:
Since it was replaced by mohawk.
Fixes #54.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,10 +97,6 @@ The REST API methods are documented on
     index.listNamespaces('mozilla-central', payload={'continuationToken': 'a_token'})
     ```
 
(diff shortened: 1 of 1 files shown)


gregarndt:
John, this should take care of Ed's request to cherry pick this commit back into master

PR #71, opened 2017-05-11
fix scriptworker

escapewindow:
This patch:

- removes the `session.close()` on `__del__`.  The `session.close()`
removed spurious errors in tools that run for a small period of time,
but causes issues for long-running servers like scriptworker that
generate a long-lived session.  We can probably stop defining sessions in scriptworker and defaulting to the sessions defined by taskcluster, but this is a fairly large change that needs to propagate through {signing,beetmover,pushapk}script.

- Defines `rerr` in `AsyncBaseClient._makeHttpRequest`.  I saw some
error messages about referencing `rerr` before it was defined during
scriptworker testing.  We can either set it to `None` or to a string;
setting it to a string seemed like the most expedient path forward.

I ran this through scriptworker tests against both the master branch and the claimWork branch, and this patch gets the integration tests green.

@jhford, can you take a look and spin a new release if it looks good?


taskcluster/async/asyncclient.py
```diff
@@ -152,6 +147,7 @@ def __del__(self):
             status = response.status
             if status == 204:
                 return None
+            rerr = await response.text()
```
jhford: Can we rename this variable here and where it's used later in the function?  We're using `rerr` elsewhere in this function for exceptions and here, it's a string.

jhford: Actually, let's match the sync class and do  `log.warn('Retrying because of a %s status code' % status)`

jhford:
`Submitting dist/taskcluster-1.3.1.tar.gz to https://upload.pypi.org/legacy/`

Released

PR #73, opened 2017-05-15
more informative error messages

### LINKED ISSUE #72: tc.py http errors lose context
I'm hitting an "Unknown Server Error" with `claimWork` calls.  Due to the current [`TaskclusterRestFailure` block](https://github.com/taskcluster/taskcluster-client.py/blob/master/taskcluster/async/asyncclient.py#L166-L188), the message only contains 'Unknown Server Error', with no status or message from the server.

I'm not sure how we're supposed to deal with this.  I'd love for shorter messages to be embedded into the message, or at least the status code.  Otherwise we'll have to make sure to catch `TaskclusterRestFailure` to inspect the body.

12:01 <•jonasfj> aki: I think it's queue side issues..
12:01 <aki> ok
12:02 <•jonasfj> aki: if you can print more information from the exception like the body that would be interesting...
12:02 <aki> should i catch those?
12:02 <•jonasfj> probably...
12:02 <•jonasfj> if polling fails... try again... :)
12:02 <aki> i think taskcluster-client.py is eating the error, if any: https://github.com/taskcluster/taskcluster-client.py/blob/master/taskcluster/async/asyncclient.py#L166
12:02 <•jonasfj> I'm not sure we have a quick solution for this...
12:03 <•jonasfj> yeah, the python client isn't good a propogating error information..
12:03 <•jonasfj> the node one attaches pretty much everthing to the error object
12:04 <aki> we might want to get the data into the message
12:04 <•dustin> open an issue?
12:04 <•dustin> I think this isn't the first time we've bumped into this :(
12:05 <•jonasfj> aki: in node I always log errors with all the properties... that helps... messages otherwise risk becoming way too bug..
12:05 <•jonasfj> *big

escapewindow:
I think the real fix here is the `await response.json()`... we were getting a coroutine in `data`, which is not a dict, so we skipped the `message = data.get('message')` line.

I added some more info to the generic "Unknown Server Error" message.  I think this will help, though I don't know what scenarios we'll hit this, so I'm less certain about that change.

Fixes #72 .


taskcluster/async/asyncclient.py
```diff
@@ -171,6 +171,8 @@ def _createSession(self):
                         message = "Authentication Error"
                     elif status == 500:
                         message = "Internal Server Error"
+                    else:
+                        message = "Unknown Server Error %s\n%s" % (str(status), data)
```
jhford: I'd rather that we cap this at, say, 1024 characters.  My concern is that an extremely large response happens to parse as valid JSON data (so as not to raise in the underlying `json.loads`) and utterly spams stdout.

PR #76, opened 2017-06-30
Bug 1401984 - Allow query strings in the API

srfraser:
With the restrictions previously in place it wasn't possible to
use continuationToken, and other query string parameters.

Adjust the argument processing so that it uses the 'query' option
in 'funcinfo' and if keyword arguments are used, populate both the
arg list (which is still only used to fill out the path in the url)
and a query string.


jhford:
Hey!  Thanks for the PR.  I'm a little concerned that the querystring arguments are being passed in in the same place as the non-query arguments.  This could become an issue when we have querystring arguments which have the same name as route-based parameters.  I was thinking we could have it so that example signatures for methods would be:

```python
# without querystring 
client.method(payload, k=v, k2=v2)
client.method(v, v2, payload)

# with querystring 
client.method(payload, queryopts, k=v, k2=v2)
client.method(v, v2, payload, queryopts)
```

For the `*args` we can check if the method supports query string options, then determine if the last two arguments are `dict` and use query string logic if so.  For the `**kwargs`, we can check if `*args[1]` is a `dict` or not... maybe...  that might need more thought as the first kwarg might be caught as the second positional argument.

srfraser:
I hadn't considered the duplicate name situation! If the parameters have the same name, wouldn't they have the same value? I'm trying to imagine a situation where they're different and it's not confusing to the users.

There's an increasing number of combinations going with the options to pass in arguments, and I find it hard to tell which ones will work.

I like the idea of `queryopts` as a `dict`, but it does mean that the person using the python module has to understand that they're used differently in the API, when you could argue that from the client's point of view they are just parameters and they shouldn't have to know.

srfraser:
This will become increasingly useful for us, so I'd like to revisit this.

I'm not keen on the idea that the person using the python module needs to know whether something ends up in the query arguments or the route, I think that's a level of knowledge that shouldn't be required, although if it's the way it's preferred, we can rework it. 

How likely are duplicate names?

jhford:
It's not strictly a question about likelihood.  It's possible, and until we are certain that it's not, we need to support the case that there are duplicates.  I'm going to take a look today to see if it'd be possible to support a small change to the method signature to support query string arguments without breaking the existing api.

Regarding callers needing to know about query arguments vs those in the route, I'd say that it's actually an important thing to know the difference between.  The docs describe them differently, they are used to represent different things and the canonical taskcluster client, the js one, also requires them to be specified in a different place.  What I would agree on is that the existence of query string arguments at all is definitely something we should have a discussion on.  The uses for query string arguments are part of request bodies in other APIs.

Out of curiousity, do you use the *args or **kwargs to specify arguments when you use the client?

jhford:
Closing in favour of https://github.com/taskcluster/taskcluster-client.py/pull/86

PR #77, opened 2017-07-12
Bug 1378232: months and years are 30 and 365 days, respectively

djmitche:
This is the same set of tests I just added to the JS client, and comes from JSON-e.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ tests_require = [
     'pgpy>=0.4.0,<0.5',
     'tox==2.3.2',
     'coverage==4.1b2',
+    'python-dateutil==2.6.0',
 ]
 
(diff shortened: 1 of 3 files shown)


djmitche:
The TC failure is #78.  Turns out `yarn test` isn't a great way to run Python tests.

djmitche:
I think the other alternative (increment the year or day) is more fuzzy, what with leap seconds and leap days and months with different numbers of days and so on.  This approach is at least easy to understand, and I think adequate for the uses to which the function is generally put.  In fact it hasn't caused any issues that all three fromNow implementations did different things with months and years -- the inconsistency just bugged me :)

But at least for JSON-e, we should document this..

PR #82, opened 2017-08-07
Only include taskcluster.async on Python 3

edmorley:
Since it doesn't work on Python 2, so might as well be excluded there from the wheel and/or files installed from the sdist.

PR #83, opened 2017-08-08
add more taskcluster.async documentation

### LINKED ISSUE #81: Document taskcluster vs taskcluster.async in the README
On Python 3 only, there's an asynchronous API available in the form of `taskcluster.async`.

However the readme currently doesn't explicitly state that it exists (it's just mentioned in passing in some examples), and that it's Python 3 only.

(On that note, it's probably worth also mentioning in the readme that the minimum Python 3 version is 3.5)

escapewindow:
Fixes #81 .


README.md
```diff
@@ -44,6 +44,28 @@ but unnamed temporary credentials can be created regardless of your scopes.## AP
 The REST API methods are documented on
 [http://docs.taskcluster.net/](http://docs.taskcluster.net/)
 
+## Sync vs Async
+
+The objects under `taskcluster` (e.g., `taskcluster.Queue`) are python2-compatible and operate synchronously. This allows for taskcluster api access, though for bulk operations will be considerably slower... more than 10x slower in cases where you need to run many api calls that would benefit from concurrency.
+
+The objects under `taskcluster.async` (e.g., `taskcluster.async.Queue`) require `python>=3.5`. `taskcluster.async` will only be available under the python3 wheel or the source distribution; the python2 wheel will not include it. The async objects use asyncio coroutines for concurrency; this allows us to put I/O operations in the background, so operations that require the cpu can happen sooner. The code would look something like
```
edmorley: The "`taskcluster.async` will only be available under the python3 wheel or the source distribution; the python2 wheel will not include it." bit is now no longer the case, since it was going to be a bit of a hack in #82.

README.md
```diff
@@ -44,6 +44,28 @@ but unnamed temporary credentials can be created regardless of your scopes.## AP
 The REST API methods are documented on
 [http://docs.taskcluster.net/](http://docs.taskcluster.net/)
 
+## Sync vs Async
+
+The objects under `taskcluster` (e.g., `taskcluster.Queue`) are python2-compatible and operate synchronously. This allows for taskcluster api access, though for bulk operations will be considerably slower... more than 10x slower in cases where you need to run many api calls that would benefit from concurrency.
```
jhford: Rather than the sync api is slower inherently, it's more the case that the async api can easily run with concurrency.  As it's written, it almost sounds like a single sync call is up to 10x slower than a single async call.  Let's frame it more like "This allows for making concurrent API calls, which can considerably improve performance for applicable workloads"

escapewindow: It's true it's just as fast on a single sync call. However, anyone using taskcluster-client.py against even medium-sized graphs should see speedups using async. But agreed, I can reword and make it a positive statement for async rather than a negative for sync.

escapewindow: Done.

escapewindow:
@jhford how does it look?

edmorley:
Many thanks!

PR #86, opened 2017-10-16
Query String Support

jhford:
Here's the initial work on supporting query strings.  This PR has the following changes:

1. Deterministic ordering for api entry insertion in the code generation logic
2. Moves all the argument processing into _processArgs which simplifies the sync and async makeApiCall functions, which means more shared code between async and sync
3. Adds support for building URLs and signed URLs for query-string params
4. Adds supprt for making api calls with query strings when _makeApiCall is called with appropriate arguments

Still to do:

1. Add tests for query-string handling
2. Change functionality of _processArgs to support query-string parameters in the call itself
3. Figure out why python 2.7 unit tests just plain don't work for me on my machine.  All development done with Python 3.6
4. Make sure that topic exchanges are also generated in a deterministic fashion 

Please ignore the changes to files other than taskcluster/client.py, taskcluster/async/asyncclient.py and genCode.py.

### DIFF
--- a/genCode.py
+++ b/genCode.py
@@ -81,7 +81,7 @@ def createStaticClient(name, api, genAsync=False):
     for opt in copiedOptions:
         if api.get(opt):
             lines.append('        "%s": "%s"' % (opt, api[opt]))
-    
+
     lines.extend(['    }', ''])
 
     # We need to build up some information about how the functions work
@@ -115,7 +115,7 @@ def createStaticClient(name, api, genAsync=False):
                 if entry.get('input'):
                     ds = '%s\n\nThis method takes input: ``%s``' % (ds, entry['input'])
                 if entry.get('output'):
-                    ds = '%s\n\nThis method takes output: ``%s``' % (ds, entry['output'])
+                    ds = '%s\n\nThis method gives output: ``%s``' % (ds, entry['output'])
                 if entry.get('stability'):
                     ds = '%s\n\nThis method is ``%s``' % (ds, entry['stability'])
 
@@ -165,19 +165,32 @@ def createStaticClient(name, api, genAsync=False):
                 for key in entry['routingKey']:
                     ds += '\n\n * %s: %s%s' % (key.get('name'), key.get('summary', ''), ' (required)' if key['required'] else '')
 
-                lines.append(cleanDocstring(ds, indent=4))
-
-            lines.extend([
-                '    def %s(self, *args, **kwargs):' % entry['name'],
-                '        return self._makeTopicExchange(%s, *args, **kwargs)' % repr(exRef),
-                ''
-            ])
-
+            lines.append('    def %s(self, *args, **kwargs):' % entry['name'])
+            lines.append(cleanDocstring(ds, indent=8))
+            lines.append('        ref = {')
+            for refK, refV in sorted(six.iteritems(exRef)):
+                if refK == 'routingKey':
+                    lines.append('            \'routingKey\': [')
+                    for routingKey in refV:
+                        lines.append('                {')
+                        for routingK, routingV in sorted(six.iteritems(routingKey)):
+                            if routingK in ('name', 'constant', 'multipleWords'): 
+                                lines.append('                    \'%s\': %s,' % (routingK, pprint.pformat(routingV)))
+                        lines.append('                },')
+                    lines.append('            ],')
+                else:
+                    lines.append('            \'%s\': %s,' % (refK, pprint.pformat(refV)))
+            lines.append('        }')
+            lines.append('        return self._makeTopicExchange(ref, *args, **kwargs)')
+            lines.append('')
 
     lines.append('    funcinfo = {')
-    items = functionInfo.iteritems() if six.PY2 else functionInfo.items()
-    for funcname, ref in sorted(items):
-        lines.append('        "%s": %s,' % (funcname, pprint.pformat(ref, indent=12)))
+
+    for funcname, ref in sorted(six.iteritems(functionInfo)):
+        lines.append('        "%s": {' % funcname)
+        for keyname, keyvalue in sorted(six.iteritems(ref)):
+            lines.append('            \'%s\': %s,' % (keyname, pprint.pformat(keyvalue)))
+        lines.append('        },')
     lines.append('    }')
 
     lines.extend([
@@ -192,7 +205,7 @@ def createStaticClient(name, api, genAsync=False):
             name
         ]),
         '',
-    ]) 
+    ])
 
     # Join the lines, then re-split them because some embedded new lines need
     # to be addressed
@@ -204,7 +217,7 @@ def createStaticClient(name, api, genAsync=False):
 
     # Build the final string
     return '\n'.join(lines)
-    
+
 
 # List of files which we've created for the build and release system's
 # consumption
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -164,7 +164,7 @@ class BaseClient(object):
         if not entry:
             raise exceptions.TaskclusterFailure(
                 'Requested method "%s" not found in API Reference' % methodName)
-        routeParams, _, query = self._processArgs(entry, *args, **kwargs)
+        routeParams, _, query, _, _ = self._processArgs(entry, *args, **kwargs)
         route = self._subArgsInRoute(entry, routeParams)
         if query:
             route += '?' + urllib.parse.urlencode(query)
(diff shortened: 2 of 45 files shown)


genCode.py
```diff
@@ -175,9 +175,12 @@ def createStaticClient(name, api, genAsync=False):
 
 
     lines.append('    funcinfo = {')
-    items = functionInfo.iteritems() if six.PY2 else functionInfo.items()
-    for funcname, ref in sorted(items):
-        lines.append('        "%s": %s,' % (funcname, pprint.pformat(ref, indent=12)))
+
+    for funcname, ref in sorted(six.iteritems(functionInfo)):
```
jonasfj: This looks sane, but would have been nice if auto-generated code changes were in a separate commit :)

jhford: yah.... sorry about that.

gregarndt:
@escapewindow Hey Aki, mind checking this out? This is the implementation that you should be use for your wrapper script that you might write.

jhford:
@jonasfj I don't know how to redo the review request, so I'll ping you and try to figure it out.  Here's an example of how these changes will work
```python
jhford-macbook:~/taskcluster/taskcluster-client.py $ ./.tox/py35/bin/python test-cont-tok.py
Response 0 gave us 1000 more tasks
Response 1 gave us 1000 more tasks
Response 2 gave us 1000 more tasks
Response 3 gave us 1000 more tasks
Task Group JzTGxwxhQ76_Tt1dxkaG5g has 3982 tasks
jhford-macbook:~/taskcluster/taskcluster-client.py $ cat test-cont-tok.py
import taskcluster
queue = taskcluster.Queue()
i = 0
tasks = 0
outcome = queue.listTaskGroup('JzTGxwxhQ76_Tt1dxkaG5g')
while outcome.get('continuationToken'):
    print('Response %d gave us %d more tasks' % (i, len(outcome['tasks'])))
    if outcome.get('continuationToken'):
        outcome = queue.listTaskGroup('JzTGxwxhQ76_Tt1dxkaG5g', query={'continuationToken': outcome.get('continuationToken')})
    i += 1
    tasks += len(outcome.get('tasks', []))
print('Task Group %s has %d tasks' % (outcome['taskGroupId'], tasks))

```

A couple notes:

1. We're still supporting the two old versions of the calling convention, through some checks.  Please let me know if there are any edge cases you think the checks wouldn't cover.  `test/test_client.py:TestProcessArgs` is the relevant test case, `client.py:BaseClient._processArgs()` is the relevant function
2. I'd like to deprecate the `method(payload, k1=v1)` style of calling methods but I'm not totally convinced whether or not to do it.  Basically, this format is only allowed when all the kwargs are exactly the same as the api args, but doesn't support query strings, so it should be safe.
3. I was thinking of supporting a `paginationHandler` function which can be passed in as a kwarg using the `method(query={}, payload={})` form which gets called on each response.
4. I reverted all the code-generation stuff, we'll need to do an api update before releasing
5. I won't push to pypi for a little while to let people test, since semver isn't a python standard

jhford:
@srfraser do you think you could try out this branch with your tools?

taskcluster/client.py
```diff
@@ -219,62 +222,125 @@ def genBewit():
...
         return self._makeHttpRequest(entry['method'], route, payload)
 
-    def _processArgs(self, entry, *args, **kwargs):
-        """ Take the list of required arguments, positional arguments
-        and keyword arguments and return a dictionary which maps the
-        value of the given arguments to the required parameters.
-
-        Keyword arguments will overwrite positional arguments.
+    def _processArgs(self, entry, *_args, **_kwargs):
+        """ Figure out, given an entry, positional and keyword arguments, what
```
imbstack: Tiny Nit: I think this reads a bit better as "Given an entry, positional and keyword arguments, figure out what the query-string options, payload and api arguments are."

taskcluster/client.py
```diff
@@ -219,62 +222,125 @@ def genBewit():
...
+        the query-string options, payload and api arguments are.
         """
 
+        # We need the args to be a list so we can mutate them
+        args = list(_args)
+        kwargs = copy.deepcopy(_kwargs)
+        log.debug(args)
```
imbstack: If these are being left in, perhaps they should be something like  "${entry} called with args ${args} and kwargs ${kwargs}". Not that it has to be exactly that, but it might be a bit nicer.

It is also possible that I don't remember how python logging works and this is just fine as it is.

taskcluster/client.py
```diff
@@ -219,62 +222,125 @@ def genBewit():
...
         reqArgs = entry['args']
-        data = {}
+        routeParams = {}
+
+        query = {}
+        payload = None
+        kwApiArgs = {}
+
+        # There are three formats for calling methods:
+        #   1. method(v1, v1, payload)
```
jonasfj: Do this mode not support query params?

Note: I ask because this is the mode we use in all other client libraries.
Typically, it looks like: `service.method(v1, v2, payload, query)` where `query` is optional, and `payload` is required if a payload is accepted by the method.

jhford: it doesn't, and it cannot.  In other languages that's simple because you know how many position arguments are required, and for sure if payload is expected, so you can make the last arg optional.  In the python client, this is complicated by supportingn bare kwargs and args in the same call.

I think it's even better to force `query={}` because it just makes it so much clearer what's going on.

jhford:
Not sure when the commit is going to show up here, github's cache invalidation is busted right now, but I pushed a change to this branch that supports:
```
jhford-macbook:~/taskcluster/taskcluster-client.py $ cat pagination-example.py
import taskcluster
queue = taskcluster.Queue()

responses = []

def x(y):
    print("%d tasks fetched" % len(y.get('tasks', [])))
    responses.append(y)

queue.listTaskGroup('JzTGxwxhQ76_Tt1dxkaG5g', paginationHandler=x)

tasks = 0
for response in responses:
    tasks += len(response.get('tasks', []))

print("%d requests fetch %d tasks" % (len(responses), tasks))

jhford-macbook:~/taskcluster/taskcluster-client.py $ ./.tox/py35/bin/python pagination-example.py
1000 tasks fetched
1000 tasks fetched
1000 tasks fetched
1000 tasks fetched
982 tasks fetched
5 requests fetch 4982 tasks
```

taskcluster/client.py
jhford: Agreed

taskcluster/client.py
jhford: they're not supposed to be left in, but I couldn't for the life of me remember where i'd typed them.  i'll remove them

escapewindow:
This adds the paginationHandler to asyncclient:

```diff
diff --git a/taskcluster/async/asyncclient.py b/taskcluster/async/asyncclient.py
index 57362d3..82af6dd 100644
--- a/taskcluster/async/asyncclient.py
+++ b/taskcluster/async/asyncclient.py
@@ -4,7 +4,6 @@ from __future__ import absolute_import, division, print_function

 import os
 import logging
-import copy
 import hashlib
 import hmac
 import datetime
@@ -60,11 +59,29 @@ class AsyncBaseClient(BaseClient):
         """ This function is used to dispatch calls to other functions
         for a given API Reference entry"""

-        routeParams, payload, query, _, _ = self._processArgs(entry, *args, **kwargs)
+        x = self._processArgs(entry, *args, **kwargs)
+        routeParams, payload, query, paginationHandler, paginationLimit = x
         route = self._subArgsInRoute(entry, routeParams)
-        log.debug('Route is: %s', route)

-        return await self._makeHttpRequest(entry['method'], route, payload)
+        # TODO: Check for limit being in the Query of the api ref
+        if paginationLimit and 'limit' in entry.get('query', []):
+            query['limit'] = paginationLimit
+
+        if query:
+            _route = route + '?' + urllib.parse.urlencode(query)
+        else:
+            _route = route
+        response = await self._makeHttpRequest(entry['method'], _route, payload)
+
+        if paginationHandler:
+            paginationHandler(response)
+            while response.get('continuationToken'):
+                query['continuationToken'] = response['continuationToken']
+                _route = route + '?' + urllib.parse.urlencode(query)
+                response = await self._makeHttpRequest(entry['method'], _route, payload)
+                paginationHandler(response)
+        else:
+            return response

     async def _makeHttpRequest(self, method, route, payload):
         """ Make an HTTP Request for the API endpoint.  This method wraps
```

This is a test script that uses it:

```python
#!/usr/bin/env python
import asyncio
import pprint


def main():
    task_group = 'JzTGxwxhQ76_Tt1dxkaG5g'
    output = []
    queue = Queue()

    def handler(contents):
        tasks = contents['tasks']
        print("handler called! adding {} tasks".format(len(tasks)))
        output.extend(contents['tasks'])

    loop.run_until_complete(
        queue.listTaskGroup('JzTGxwxhQ76_Tt1dxkaG5g', paginationHandler=handler)
    )
    print('Task Group %s has %d tasks' % (task_group, len(output)))
    pprint.pprint(output)


__name__ == '__main__' and main()
```

jhford:
Merged to master, thanks everyone!

Issue #87, opened 2017-10-19
Check and update dependencies


edmorley:
Continuing from:
https://github.com/taskcluster/taskcluster-client.py/commit/90cdde8f0fb05d6d8b7937e3f255e2813df4183c#commitcomment-25074883

I copied out the version lists from setup.py, trimmed the quotes and commas, which gives roughly:
```
# tests_require
nose==1.3.7
nose-exclude==0.5.0
httmock==1.2.6
rednose==1.2.1
mock==1.0.1
setuptools-lint==0.3
flake8==2.5.0
psutil==2.1.3
hypothesis==3.6.1
pgpy>=0.4.0,<0.5
tox==2.3.2
coverage==4.1b2
python-dateutil==2.6.0

# install_requires
requests>=2.4.3,<3
mohawk>=0.3.4,<0.4
slugid>=1.0.7,<2
six>=1.10.0,<1.11

# python 2 tests_require addition
subprocess32==3.2.6

# python 3 install_requires addition
aiohttp>=2.0.0,<3
async_timeout>=1.2.0,<2
```

And then pasted that in https://pyup.io/tools/requirements-checker/ , which shows updates for a few things.

It turns out there is an issue open for adding read-only setup.py support to pyup.io:
https://github.com/pyupio/pyup/issues/137

And there's in fact a workaround suggested there for the moment:
https://github.com/pyupio/pyup/issues/137#issuecomment-323789520

jhford:
but if we did that, wouldn't we be forcing installation of subprocess32 on py3 and async_timeout/aiohttp on py2?

edmorley:
The above isn't a proposed list of dependencies to use, it's what I extracted from setup.py for easy pasting in the pyup online checker tool.

jhford:
ahhhh, i was thinking the idea was to commit that as a requirements.txt file, which we can't do because of the different requirements for python 2 and 3.

Thanks!

escapewindow:
Per [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1531021#c7), we're thinking that:

- since `taskcluster-client.py` infrequently updates its deps, and
- since a number of downstream users use pyup and similar to regularly update dependencies,

it may be beneficial to stop pinning deps in `taskcluster-client.py`, and push that responsibility down to the users. An alternate solution could involve regularly using pyup or similar to bump dependencies for `taskcluster-client.py`, run the tests, and land the PR+release tc.py if the tests pass.

[...]

PR #89, opened 2017-11-15
Update install_requires to allow for newer async_timeout

edmorley:
The breaking change made by `async_timeout` doesn't affect this project, since the only usage is with a non-zero timeout:
https://github.com/aio-libs/async-timeout/blob/master/CHANGES.rst#200-2017-10-09
https://github.com/taskcluster/taskcluster-client.py/blob/888c98f8471ee4ce9ae4445353e99ef1560ec0c4/taskcluster/async/asyncutils.py#L39

~The `six` dependency update has been facilitated by switching to PGPy master, since I'd prefer to not wait any longer for a new release. This can be restored to a PyPI version once a release occurs.~

Refs #87 (but doesn't close it, since there's more work to do).

~This PR unblocks two Treeherder PRs:
mozilla/treeherder#2814
mozilla/treeherder#2891 (will be updated once a new taskcluster client is released)~


edmorley:
Eugh, setuptools doesn't support the full range of version specifiers in `tests_require`, so a URL can't be used (unless using `dependency_links`, but I think that has side effects even for non-`develop` installs, so probably not worth doing).

The only other solution would be to move `tests_require` to a pip requirements file, and having that used as part of the tox run (this is probably worth doing anyway, to help with being able to use pyup.io in #87 - and would help the testing setup in this repo become a bit more standard compared to other projects). 

However I don't really have the time for all that right now, so I guess I'll just have to hold off until the next PGPy release is made.

jhford:
I agree, we should wait for the release to happen.  I'd rather not depend on a specific commit of a repo in the main release.  Are we able to update async timeout independently?  I'd be happy to update that if it works.

edmorley:
I've revised the PR to just update `async_timeout` only :-)

(The restricted version range for async timeout actually caused gerard-majax breakage the other day due to pip's resolver non-determinism, so would be good to relax it)

edmorley:
@jhford could this be merged?

jhford:
Sorry!  Didn't see your new commit.  Looks good to me

edmorley:
Many thanks :-)

PR #90, opened 2017-11-15
Include linting in tox

### LINKED ISSUE #88: Add lint/setupcheck tests to Travis run
The `test.sh` script does linting and a few other things, however it's not run on Travis (and is failing for me locally under MSYS2), so hard to figure out if updating dependencies has broken anything (in #87).

Ideally those steps would be defined in tox.ini and run via tox both locally and on travis.

jhford:
Basically, I'd like to move as much as I can into tox.ini.  This should resolve #88 and also do a bit of general cleanup that's overdue.


jhford:
@edmorley fyi

djmitche:
Is this ready for review?  I'm not sure what is being ignored..

jhford:
Please ignore README.md and genDocs.py, i will fix those.  it's probably related to how I'm handling utf-8 encoding and decoding.  Otherwise, the rest of the code is ready for review.

taskcluster/async/scheduler.py
```diff
@@ -1,293 +0,0 @@
-# coding=utf-8
```
djmitche: And, it looks like, remove scheduler support?  That should probably be broken out into a separate commit so it can be seen in the git log.

Issue #91, opened 2017-11-19
Make it easier to use the Taskcluster Proxy


indygreg:
As best I can tell, in order to use Taskcluster Proxy (https://docs.taskcluster.net/reference/workers/docker-worker/docs/features#feature-taskclusterproxy-), you need to explicitly define the URL endpoint for each API client you are using. e.g. `queue = Queue({'baseUrl': 'http://taskcluster/queue/v1'}, session=session)`.

I think it would be rad if you could pass a `tc_proxy=True` (or similar) somewhere once, and all API clients would automatically use the Taskcluster Proxy.

jhford:
that sounds reasonable!

PR #92, opened 2017-11-21
Support easier taskcluster proxy integration

jhford:


jhford:
@indygreg: What do you think about this for fixing #91?  Basically, you specify `tcProxy` to a truthy value and it'll automatically set tc-proxy urls.  Example: `https://queue.taskcluster.net/v1/task` becomes `http://taskcluster/queue/v1/task`.  I've done things to handle the unspecified case of a service like `queue.b.taskcluster` by setting the proxied url to `taskcluster/queue.b`

gregarndt:
What if instead of a boolean, it's a proxy address to use?  I only suggest this because while http://taskcluster is the default for docker-worker, I do not believe that will be the case for taskcluster-worker and the new vision for the proxy.  @jonasfj can correct me if I'm wrong.

jhford:
at that point, just setting `baseUrl=http://taskcluster/queue` would probably be sufficient

indygreg:
I'm happy to set a base URL. But only once. I don't want to have to type "http://taskcluster/queue" "http://taskcluster/index" etc.

jonasfj:
In tc-worker we support worker plugins  which can expose a proxy to the sandbox.
This means that the `baseURL` will be something like `http://taskcluster/proxy/queue.taskcluster.net/v1`.

I agree with gps we don't want to have to set this multiple times. And that makes me think this might tie in a bit with some of the thoughts we are having on re-deployability, as this will make client library require a `baseDomain` (or something) along with `clientId`, `accessToken`.

Maybe the proxy is a different `baseDomain`, and maybe it's not called `baseDomain` :)
Maybe it's called `domain` and by convention services are available at `domain/<service>`...

jhford:
do we not use taskcluster-proxy in tc worker?

maybe this is something we should codify in taskcluster-lib-api and remove the variability?  we could have a 'baseUrl' and a 'proxyUrl'.  The former would remain as is, and the latter would be used by the proxy to know where to serve the proxy and clients to know where to consume from?  We could at the same time set an stanardard that environment variable `TASKCLUSTER_PROXY=true` would trigger clients to automatically use the TC proxy.

jonasfj:
> do we not use taskcluster-proxy in tc worker?

No the same implementation no... We try to keep the worker easy to deploy by making everything as single go binary (well, not with QEMU engine), but other engines. This also makes it easier to test things.

It's the same concept as taskcluster-proxy, just more abstract implementation.

petemoore:
This topic got me thinking a lot - so here is some bike shedding.

It occurred to me that the concept of an http(s) proxy is something built into the http specification itself, and for good reason. The concerns we are iterating through above (and others we've not yet touched, such as how to rewrite urls if an upstream service is running on a non-standard port) would have standard solutions, if we made taskcluster-proxy an actual http proxy. This would allow standard http clients to also interact with the proxy (e.g. `curl -x`) even if they were calling endpoints that the proxy doesn't know about, since the proxy url is independent from the request url, so requests can be made to services the proxy doesn't know about, without needing to update/configure the proxy itself.

The downside of using an http proxy is it can be less transparent that a proxy sits in the middle, or even worse, that you don't realise it is injecting `Authorization` headers into your requests. Avoiding an http proxy solves this problem, but does require that the taskcluster proxy has careful url rewriting making sure to cover all possible target urls (including non-standard schemes, ports, user info etc - not just the url path). One alternative to encoding this information in the url path would be to make the full target url a query string parameter. In this case, the proxy would be a global url configured in the taskcluster client (non-empty string), and when set, the client would query this url, appending the target service url as a query string parameter to the request, which the proxy would then read, add an `Authorization` header, and call.

The disadvantage of encoding the url as a query string parameter, is it requires a backwardly-incompatible change to the proxy which would impact many tasks. The advantage is that the target url is transparent, and arbitrary, not needing to worry about encoding all parts of the url such as scheme, port, user info into it.

In any case, I think setting a boolean for the proxy might be little bit too simplistic to cover some scenarios, so perhaps a proxy url base like `http://taskcluster` as a string would support more use cases, e.g. running outside of docker worker linked containers. That might provide a good balance of the above concerns and still support the simple docker-worker use case.

At some point I would like to think we'll have a `~/.taskcluster/config` file that will allow users to predefine endpoints, proxies, authentication data, etc - and that our command line tools will respect this config, in addition to environment variables that they typically already support.

But I really like the idea that tasks start using a taskcluster-client, with proxy support built in, rather than calling raw urls - so this is a great move forward.

jonasfj:
@petemoore, there are quite a few solutions to this... they all have ups and down...
A generic HTTP proxy can't easily compute and sign the content-hash. Maybe open an RFC for figuring this out.

PR #93, opened 2017-12-20
Remove clock skew in temp credentials generation

petemoore:
This is now taken care of in the auth service directly:
https://github.com/taskcluster/taskcluster-auth/pull/117

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,9 @@ temporary credentials as follows. Notice that the credentials cannot last more
 than 31 days, and you can only revoke them by revoking the credentials that was
 used to issue them (this takes up to one hour).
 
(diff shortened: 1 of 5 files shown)


jhford:
looks good to me!  I'm curious why only python3.6 fails, but I'll retrigger that when 117 lands.

petemoore:
> looks good to me! I'm curious why only python3.6 fails, but I'll retrigger that when 117 lands.

I think it was just a case of luck, depending on which worker got the travis job, and whether its clock was ahead or behind the auth service dyno's clock that serviced the requests since at the time it failed it was running against an auth service that didn't apply clock skew. I retriggered tests after landing taskcluster/taskcluster-auth#117 and now it passes, so I think we're safe to merge.

PR #94, opened 2018-01-23
Relax six requirement

edmorley:
So consumers of taskcluster-py aren't prevented from updating.

The newer `six` requires a fix to PGPy that has not yet been released, so we have to manually update to a newer revision.


edmorley:
@jhford with the manual update tests now pass on newer six. Once this is merged I'll open an issue to track removing the zip archive install from `devDeps.py`.

edmorley:
@djmitche don't suppose you'd mind reviewing?

Since the PR was opened there still hasn't been any update on a PGPy release (eg SecurityInnovation/PGPy/issues/236), so not blocking on it seems like the best course of action at the moment (especially given that it's testing only).

djmitche:
The patch looks simple enough, but you and @jhford were involved in this from the beginning so I'll defer to him for review.

setup.py
```diff
@@ -19,6 +19,7 @@
     'flake8==2.5.0',
     'psutil==2.1.3',
     'hypothesis==3.6.1',
+    # Note this is temporarily overridden in devDep.py.
     'pgpy>=0.4.0,<0.5',
```
jhford: We shouldn't have a set of version specs in setup.py that's known to not work together.  That means either requiring greater than pgpy 0.4.3 or six less than 1.11.  Since we're not sure when/if pgpy 0.4.4 will be released, and whether it'll included the fix, requiring six less than 1.11 is the safer of the two.

edmorley: This is a test only dependency though...

jhford: pgpy is a runtime dependency as well [1].  Even though we have a failsafe for having a completely missing pgpy, we don't have any safety for finding pgpy but getting a broken version.  The problem here is that with six 1.11 and pgpy 0.4.3, someone could have production breakage and setup.py would allow that installation to occur.


[1] https://github.com/taskcluster/taskcluster-client.py/blob/f2ebe24a9adaf99b57c5ca26ac05416b5195aaa2/taskcluster/utils.py#L29

edmorley: The root cause here is that ideally taskcluster-client.py shouldn't be using test_requires at all, and instead using a requirements file / tox based solution for testing (like most other packages do now).

However I don't have the time to spare to do that at the moment, and given this is a test only dependency, I don't really see the harm in making sure the test environment installs the updated version that's no longer buggy with modern six. Holding back other consumers of the package from keeping their deps (including six) up to date is worse IMO.

edmorley: > pgpy is a runtime dependency as well

Oh right. In that case, perhaps it would be clearer to list is under `extras_require` to make it clearer it's not just a test dep?

edmorley:
Treeherder no longer uses this package (the Taskcluster auth parts now happen client side using taskcluster-client-web). I was hoping to clear this out of my open GitHub PRs list (which always seems to have a dozen or so stragglers from less active projects), but it seems like this isn't as simple as it looked at first glance - so closing out for now :-)

setup.py
jhford: Another option would be to split out the encrypted environment variable support into another library and update this library independently.

djmitche: Oh, is that what this is for?  Encrypted env vars are deprecated and *heavily* reommended against.  We could just remove the functionality.

jhford: yep!  that sounds like a great option to me.

@rail @mozbhearsum would this be a problem for either of you?  If so, would you be opposed to splitting it out into its own library?

rail: I think this should be OK. We still use encrypted env variables and pgpy for ESR52, but the versions are pinned.

PR #95, opened 2018-02-09
give more informative error messages for api calls

jhford:
https://bugzilla.mozilla.org/show_bug.cgi?id=1435980

Let's give more informative error messages when we fail.

### DIFF
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -493,13 +493,16 @@ class BaseClient(object):
 
             # Handle non 2xx status code and retry if possible
             status = response.status_code
+
             if status == 204:
                 return None
+
+            statusMsg = response.reason
 
             # Catch retryable errors and go to the beginning of the loop
             # to do the retry
             if 500 <= status and status < 600 and retry < retries:
-                log.warn('Retrying because of a %s status code' % status)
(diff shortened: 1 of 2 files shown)


taskcluster/client.py
```diff
@@ -509,15 +512,10 @@ def _makeHttpRequest(self, method, route, payload):
...
-                message = "Unknown Server Error"
                 if isinstance(data, dict):
                     message = data.get('message')
-                else:
-                    if status == 401:
-                        message = "Authentication Error"
-                    elif status == 500:
-                        message = "Internal Server Error"
+                if not message:
+                    message = '%d: %s' % (status, statusMsg)
```
petemoore: It looks like this is causing:

```
    test/test_client.py line 384 in test_failure_status_code
      self.client._makeHttpRequest('GET', 'http://www.example.com', None)
    taskcluster/client.py line 517 in _makeHttpRequest
      if not message:
   UnboundLocalError: local variable 'message' referenced before assignment
```

jhford:
sorry about that, Pete.  I wasn't able to get the tests to work on my laptop at the time, but have fixed it on my desktop.

PR #98, opened 2018-02-22
Return None for Content-Length=0

guptatrisha97:
Checked the response length using the Content-Length header and status code and return None if it is 0. This is to resolve empty server responses.

### DIFF
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -496,6 +496,11 @@ class BaseClient(object):
             if status == 204:
                 return None
 
+            # Handle content length 0 (empty server responses)
+            servresp = response.length
+            if servresp == 0:
+                return None
+            
             # Catch retryable errors and go to the beginning of the loop
             # to do the retry
             if 500 <= status and status < 600 and retry < retries:
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -497,7 +497,7 @@ class BaseClient(object):
                 return None
 
             # Handle content length 0 (empty server responses)
-            if header("Content-Length: ".filesize(response)) == 0:
+            if headers.get('Content-Length') == 0:
                 return None
             
             # Catch retryable errors and go to the beginning of the loop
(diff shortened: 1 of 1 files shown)


taskcluster/client.py
```diff
@@ -496,6 +496,11 @@ def _makeHttpRequest(self, method, route, payload):
             if status == 204:
                 return None
 
+            # Handle content length 0 (empty server responses)
+            servresp = response.length
```
jhford: Hmm, looks like this doesn't work
```python
(py27) [jhford@jhford-desktop taskcluster-client.py]$ python
Python 2.7.14 (default, Feb 17 2018, 10:42:17) 
[GCC 7.3.1 20180130 (Red Hat 7.3.1-2)] on linux2
>>> import requests
>>> response = requests.get('http://www.google.com')
>>> response.length
  File "<stdin>", line 1, in <module>
AttributeError: 'Response' object has no attribute 'length'
>>> 
```

I think we might want to look at:
```python
response.headers['Content-Length']
```

Since it's a string, we'll need to parse it into an int.  We should handle the cases that the header isn't there, that it doesn't parse as an int and that it parses as an invalid int (e.g. negative)

taskcluster/client.py
```diff
@@ -496,6 +496,10 @@ def _makeHttpRequest(self, method, route, payload):
             if status == 204:
                 return None
 
+            # Handle content length 0 (empty server responses)
+            if header("Content-Length: ".filesize(response)) == 0:
```
jhford: Hmm, I'm not sure if this is right.  I'm not sure what's intended here.  The `header` is actually a python dictionary, so you'd either need to use subscripting, or the `.get()`/`.has()` accessors.

I'm also not sure about the `.filesize` method.  I don't see that as a method on the string type.

```
Python 2.7.10 (default, Jul 15 2017, 17:16:57)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.31)] on darwin
>>> "a-string".filesize
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute 'filesize'
```

jhford: Also, `header()` is not a defined function.
```
>>> header()
  File "<stdin>", line 1, in <module>
NameError: name 'header' is not defined
>>> 
```

PR #99, opened 2018-02-26
Checked for Non JSON Responses

guptatrisha97:
Returned None if the response given by the server was non json. Basically used content type to check if it was application/json or not.

### DIFF
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -492,10 +492,11 @@ class BaseClient(object):
                 )
                 
             # Handles Non JSON responses
-            answer = response.get_content_type()
-            typecheck = "application/json"
-            if answer != typecheck:
-                return None
+            answer = response.headers['Content-Type']
+            answer = answer.strip()
+            contentType = "application/json"
+            if answer != contentType:
+                return response.raw
 
             # Handle non 2xx status code and retry if possible
             status = response.status_code
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -490,7 +490,7 @@ class BaseClient(object):
                     "Failed to establish connection",
                     superExc=rerr
                 )
-                
+            
             # Handle non 2xx status code and retry if possible
             status = response.status_code
             if status == 204:
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -491,13 +491,6 @@ class BaseClient(object):
                     superExc=rerr
                 )
                 
-            # Handles Non JSON responses
-            answer = response.headers['Content-Type']
-            answer = answer.strip()
-            contentType = "application/json"
-            if answer != contentType:
-                return response.raw
-
             # Handle non 2xx status code and retry if possible
             status = response.status_code
             if status == 204:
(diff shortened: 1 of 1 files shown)


taskcluster/client.py
```diff
@@ -490,6 +490,12 @@ def _makeHttpRequest(self, method, route, payload):
                     "Failed to establish connection",
                     superExc=rerr
                 )
+                
+            # Handles Non JSON responses
+            answer = response.get_content_type()
```
jhford: Hmm, I'm not seeing that method in the requests library, are you able to link to the documentation?

http://docs.python-requests.org/en/master/user/quickstart/#response-headers seems to me to be a good way to get the value of the header.

We probably want to also trim whitespace from the header value before the comparison. 

We probably don't want to return `None` for all non-json responses, rather only when the status code is 204

taskcluster/client.py
```diff
@@ -490,6 +490,12 @@ def _makeHttpRequest(self, method, route, payload):
...
+            typecheck = "application/json"
```
jhford: The name is a little confusing, It's not really a check per se, but rather a value.  Maybe `contentType` is a little more clear?

taskcluster/client.py
guptatrisha97: So what do you want me to return for non-json responses?

jhford: For non-JSON responses, per [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1135584#c8) the raw `response`  value is likely the best choice.

guptatrisha97: Sure, thanks. I've changed the code to fit your review. I hope it is fine now. Requesting another review whenever you are free. Thanks!

taskcluster/client.py
```diff
@@ -490,6 +490,13 @@ def _makeHttpRequest(self, method, route, payload):
                     "Failed to establish connection",
                     superExc=rerr
                 )
+                
+            # Handles Non JSON responses
+            answer = response.headers['Content-Type']
+            answer = answer.strip()
+            contentType = "application/json"
+            if answer != contentType:
+                return response.raw
```
jhford: I didn't know that they exposed a `.raw` property here, but I think that returning the requests-library response (i.e. `return response`) should be enough.

taskcluster/client.py
```diff
@@ -490,6 +490,13 @@ def _makeHttpRequest(self, method, route, payload):
                     "Failed to establish connection",
                     superExc=rerr
                 )
+                
+            # Handles Non JSON responses
+            answer = response.headers['Content-Type']
+            answer = answer.strip()
+            contentType = "application/json"
+            if answer != contentType:
```
jhford: We can probably shorten this to 

```python
if response.headers['Content-Type'].strip() != 'application/json':
  return response
```

taskcluster/client.py
jhford: We also need to make this choice later on.  The reason is that we haven't yet classified the response as an error or a successful response.  If we did this here, we'd never be checking for non-200 series responses.  This logic should probably go near the existing line `return response.json()`.

taskcluster/client.py
```diff
@@ -535,9 +535,7 @@ def _makeHttpRequest(self, method, route, payload):
                 )
 
             # Try to load JSON
-            try:
-                return response.json()
-            except ValueError:
+            if response.headers['Content-Type'].strip() != 'application/json':
```
jhford: This is no longer parsing the body if it is `application/json`, which we need to happen.  We also still need to put this in a try/catch.

PR #100, opened 2018-03-07
Support aiohttp3

### LINKED ISSUE #96: Upgrade to aiohttp 3.x.y
From https://github.com/mozilla-releng/scriptworker/pull/184:

aiohttp 3.0.0 was [released yesterday](https://github.com/aio-libs/aiohttp/releases/tag/v3.0.0). Scriptworker has an installation problem, because taskcluster [prevents 3.0.0 from being installed](https://github.com/taskcluster/taskcluster-client.py/pull/67).

Even if we wanted taskcluster to allow 3.0.0, we would hit this [new type of failure](https://travis-ci.org/mozilla-releng/beetmoverscript/jobs/340871448#L599) on projects that depend on scriptworker.

Upgrading taskcluster would be the first step. Then, scriptworker would have to fix its test suite.

escapewindow:
* By adding `__aenter__` and `__aexit__` methods, we can allow for

  `async with Queue() as queue:` blocks.

* Currently, we're hitting a bunch of scary warnings like

        UserWarning: Creating a client session outside of coroutine is a very dangerous idea

    unless we explicitly wrap tc-client.py calls inside an

        async with aiohttp.ClientSession() as session:

    block. Let's encourage that behavior, and fall back to creating an async
    session per call otherwise.

Tested via tox, plus this simple script:

```
#!/usr/bin/env python

import aiohttp
import asyncio
import logging

logging.basicConfig()
log = logging.getLogger()
log.setLevel(logging.DEBUG)

async def foo(session=None):
    async with Queue() as queue:
        print(await queue.listProvisioners())


loop.run_until_complete(foo())
```

Fixes #96.


escapewindow:
To answer @catlee 's q:

```
#!/usr/bin/env python

import aiohttp
import asyncio
import logging

logging.basicConfig()
log = logging.getLogger()
log.setLevel(logging.DEBUG)

async def foo(session=None):
    queue = Queue()
    async with aiohttp.ClientSession() as session:
        async with Queue(session=session) as queue:
            print(await queue.listProvisioners())
        # make sure session isn't closed if we nest contextmanagers
        r = await session.get("https://google.com")
        print(await r.text())


loop.run_until_complete(foo())
```

works.

taskcluster/async/asyncclient.py
```diff
@@ -38,7 +38,9 @@
 
 def createSession(*args, **kwargs):
     """ Create a new aiohttp session.  This passes through all positional and
-    keyword arguments to the aiohttp.CreateSession() constructor
+    keyword arguments to the asyncutils.createSession() constructor.
+
+    `async with aiohttp.ClientSession(...):` in the client code is preferred.
```
catlee: ITYM `async with aiohttp.ClientSession(...) as session:`?

escapewindow:
@jhford should I flag you for review here, or would you prefer someone else review?

jhford:
I wasn't flagged for review, but I'll take a look now.

taskcluster/async/asyncclient.py
```diff
@@ -38,7 +38,10 @@
 
 def createSession(*args, **kwargs):
     """ Create a new aiohttp session.  This passes through all positional and
-    keyword arguments to the aiohttp.CreateSession() constructor
+    keyword arguments to the asyncutils.createSession() constructor.
+
+    `async with aiohttp.ClientSession(...) as session:` in the client code
```
jhford: Can you elaborate on this snippet to create a client instance and call a `ping()` method?

jhford: Oh, and I was thinking it might be nice to re-export `aiohttp.CreateSession()` from this library so that users only have to import taskcluster-client.py directly.

taskcluster/async/asyncclient.py
```diff
@@ -50,10 +53,16 @@ class AsyncBaseClient(BaseClient):
     help with this.
     """
 
+    def __init__(self, *args, **kwargs):
+        super(AsyncBaseClient, self).__init__(*args, **kwargs)
+        self._implicit = False
+        if self.session is None:
+            self._implicit = True
```
jhford: can we we make this name more explicit as to what is implict?  Maybe `_implicitSession`?

taskcluster/async/asyncclient.py
```diff
@@ -50,10 +53,16 @@ class AsyncBaseClient(BaseClient):
...
+
     def _createSession(self):
         """ Create a new aiohttp session.
         """
-        return createSession()
+        return None
```
jhford: A comment explaining why this function is stubbed for async would be useful, doesn't have to be in the docstring or anything too detailed.

taskcluster/async/asyncclient.py
escapewindow: Done.

taskcluster/async/asyncclient.py
escapewindow: Done.

taskcluster/async/asyncclient.py
escapewindow: Sure. Added a larger couple of examples. Amended `aiohttp.CreateSession` to just using `createSession`, which works as well.

escapewindow:
It looks like I'm not authorized to merge.

jhford:
thanks!

PR #103, opened 2018-04-03
Remove encrypted envs

jhford:
This also updates to the latest version of Six.  I can't run the py3.5 tests on this laptop, so let's make sure that travis passes.

@escapewindow once this merges, we should be able to do major version release.

cc @rail & @mozbhearsum for encrypted environment var support, note that this will be a major version bump

cc @edmorley, even though I know you're no longer following this client library, we're doing it :)


rail:
Yay! Less code is better :)

PR #104, opened 2018-04-10
Check for Non-json and zero content

fiennyangeln:
Done by checking if header is there, and if the key is there. Then,
do checking on the value of Content-Type and Content-Length

Bug 1135584

### DIFF
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -536,6 +536,11 @@ class BaseClient(object):
 
             # Try to load JSON
             try:
+                if hasattr(response, 'headers'):
+                    if 'application/json' not in response.headers.get('Content-Type', '').split(';'):
+                        return None
+                    if int(response.headers.get('Content-Length', len(response.content))) == 0:
+                        return None
                 return response.json()
             except ValueError:
                 return {"response": response}


taskcluster/client.py
```diff
@@ -536,9 +536,19 @@ def _makeHttpRequest(self, method, route, payload):
 
             # Try to load JSON
             try:
+                if response.headers or headers in response:
+                    if 'Content-Type' in response.headers and \
+                            'application/json' not in response.headers['Content-Type']:
```
jhford: Per the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) and [HTTP spec](https://tools.ietf.org/html/rfc7231#section-3.1.1.5), we know that the format for content-type headers are going to be something like `application/json` or `application-json; charset=utf-8`.  I think we'd have a more robust solution here if we split the header value by `;` character, and check that the trimmed first index is exactly `application/json`.  This way we know for sure that when we're parsing the response as JSON, that we're expecting a JSON value.

In an ideal world, we'd use the `charset` parameter if present to pass to the JSON parser to interpret the body correctly, but I think we make enough assumptions about UTF-8 being used elsewhere that this isn't important.

taskcluster/client.py
```diff
@@ -536,9 +536,19 @@ def _makeHttpRequest(self, method, route, payload):
 
             # Try to load JSON
             try:
+                if response.headers or headers in response:
```
jhford: This has a bit of a bug in it: `headers in response` is likely not what you want.  If there's a `headers` variable in scope here, it's almost certainly not what you're wanting to check for presence in the response object.

The Python 'repl' is really helpful in situations like this for figuring out what's going on here.  Since we know that this response object is what the request library returns from a request, we can get a sample response to do some checks against, like this:

```python
$ python
Python 2.7.14 (default, Mar 14 2018, 13:36:31) 
[GCC 7.3.1 20180303 (Red Hat 7.3.1-5)] on linux2
>>> import requests
>>> response = requests.get('http://www.google.com');
>>> response.headers
{'Content-Length': '4890', 'X-XSS-Protection': '1; mode=block', 'Content-Encoding': 'gzip', 'Set-Cookie': '1P_JAR=2018-04-10-09; expires=Thu, 10-May-2018 09:37:53 GMT; path=/; domain=.google.de, NID=127=r3tD74HByNeU2WwhWVw6T6AzkeHfem__la1lGpU_g5v3db14PheOHM0tlCMl-t_JgVR3Q7-j5TX21TiN-G-WW0wnwKog1b-jWpDi3b2lqLkuGY--Hvik1dBdkyOn4Oxs; expires=Wed, 10-Oct-2018 09:37:53 GMT; path=/; domain=.google.de; HttpOnly', 'Expires': '-1', 'Server': 'gws', 'Cache-Control': 'private, max-age=0', 'Date': 'Tue, 10 Apr 2018 09:37:53 GMT', 'P3P': 'CP="This is not a P3P policy! See g.co/p3phelp for more info."', 'Content-Type': 'text/html; charset=ISO-8859-1', 'X-Frame-Options': 'SAMEORIGIN'}
>>> headers in response
  File "<stdin>", line 1, in <module>
NameError: name 'headers' is not defined
>>> 'Content-Type' in response.headers
True
>>> 'Content-Length' in response.headers
True
>>> 
```

taskcluster/client.py
```diff
@@ -536,9 +536,19 @@ def _makeHttpRequest(self, method, route, payload):
...
+                        return None
+                    if 'Content-Length' in response.headers and \
+                            int(response.headers['Content-Length']) == 0:
+                        return None
                 return response.json()
             except ValueError:
                 return {"response": response}
+            # Error if response has no headers
+            except AttributeError:
+                return response.json()
```
jhford: Since we want to change JSON parsing to be something which we only do when we're expecting JSON response bodies, we should probably just return `{"response": response}` here.

taskcluster/client.py
fiennyangeln: I see, thats a good idea. I'll modify it according to your suggestion. thanks!!

taskcluster/client.py
fiennyangeln: ah thats true, sorry I didn't pay attention to this. I fixed it

taskcluster/client.py
fiennyangeln: okk

fiennyangeln:
Hi @jhford I modified the logic a bit, can you review it again? thank you very much!

taskcluster/client.py
```diff
@@ -536,6 +536,11 @@ def _makeHttpRequest(self, method, route, payload):
 
             # Try to load JSON
             try:
+                if hasattr(response, 'headers'):
+                    if 'application/json' not in response.headers.get('Content-Type', '').split(';'):
```
jhford: This is definitely correct to split on the `;` character and it correctly finds `application/json`, but the [spec](https://tools.ietf.org/html/rfc7231#section-3.1.1.1) only describes the mimetype as being before the first `;`.  How about making the comparison to the first index only?

fiennyangeln: Hi John, is the way to run it still the same?

```
pip install tox
git clone https://github.com/taskcluster/taskcluster-client.py
cd taskcluster-client.py
tox -e py27  # this is just to get started, eventually you should also try just "tox"
```

I got a py27 command failed thus I cant test if my change mess the test or not

jhford: yep!  that's what's needed to run it on my Fedora 28 machine.  Which OS and which pythons do you have installed?  Also, what's the error?

jhford: yep!  that's what's needed to run it on my Fedora 28 machine.  Which OS and which pythons do you have installed?  Also, what's the error?

fiennyangeln: I got this ERROR: InvocationError for command '/Users/angelinaf/Documents/taskcluster-client.py/.tox/py27/bin/python devDep.py' (exited with code 1) I use macOS. I remember previously when I open this PR i uses ubuntu and it works just fine. I will try to find why and get back to you

jhford: you could also try install python through homebrew, the system python on os x is not so great.

PR #105, opened 2018-05-10
Pinning async_timeout<3 breaks aiohttp3 support

escapewindow:
aiohttp 3 requires `async_timeout>=3,<4`

@jhford  This blocks scriptworker's [aiohttp 3 rollout](https://github.com/mozilla-releng/scriptworker/pull/200) which has been a long time coming; can we get a release shipped with this fix quickly?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ elif sys.version_info[:2] < (3, 5):
 elif sys.version_info[:2] >= (3, 5):
     install_requires.extend([
         'aiohttp>=2.0.0,<4',
-        'async_timeout>=2.0.0,<3',
+        'async_timeout>=2.0.0,<4',
     ])
 
 if __name__ == '__main__':


Callek:
devDep.py invokes pip. Pip 10.x changed:  "Move all of pip's APIs into the pip._internal package, properly reflecting the fact that pip does not currently have any public APIs. (#4696, #4700)"

escapewindow:
If we want to avoid having to release new versions of tc-client.py on version bump like this, one solution could be to not pin versions unless something's actively broken, and push dep-pinning responsibility onto the downstream modules.

jhford:
I fixed the issue with pip.  DevDep.py will try to use `pip.main` and if that fails, it'll use `pip._internal.main`.  This should fix travis.

Released 3.0.1

escapewindow:
Thanks!

PR #106, opened 2018-05-22
Add support for root urls

jhford:


taskcluster/client.py
```diff
@@ -67,6 +74,17 @@ def __init__(self, options=None, session=None):
                     except:
                         s = '%s (%s) must be unicode encodable' % (x, credentials[x])
                         raise exceptions.TaskclusterAuthFailure(s)
+
+        baseUrl = o.get('baseUrl')
+        rootUrl = o.get('rootUrl')
+
+        if rootUrl and not o.get('version'):
+            s = 'specifying a rootUrl requires a name and version in the API reference'
```
jhford: this shouldn't say `name and`

taskcluster/client.py
```diff
@@ -67,6 +74,17 @@ def __init__(self, options=None, session=None):
...
+            raise exceptions.TaskclusterFailure(s)
+        elif rootUrl:
+            name = o.get('name', urllib.parse.urlparse(baseUrl).netloc.split('.', 1)[0])
+            o['baseUrl'] = '%s/%s/v%s' % (rootUrl, name, o['version'])
```
imbstack: This probably ought to be something like `%s/api/%s/v%s` which we define in taskcluster-lib-urls. I actually just got that python implementation up for review in taskcluster/taskcluster-lib-urls#6 if you want to take a look. Once that is r+, we can use it here to build urls.

taskcluster/client.py
```diff
@@ -67,6 +74,17 @@ def __init__(self, options=None, session=None):
                     except:
                         s = '%s (%s) must be unicode encodable' % (x, credentials[x])
                         raise exceptions.TaskclusterAuthFailure(s)
+
+        baseUrl = o.get('baseUrl')
+        rootUrl = o.get('rootUrl')
+
+        if rootUrl and not o.get('version'):
```
imbstack: Unfortunately, I got the meaning of `version` in references wrong in the node implementation of this change. I'm going to have to go back and fix it but we may as well get it right here from the get-go. Apparently the `version` in a reference is just the version of the reference file format. This is discussed further in https://bugzilla.mozilla.org/show_bug.cgi?id=1460019

taskcluster/client.py
imbstack: This has now landed! https://github.com/taskcluster/taskcluster-lib-urls#python-usage

jhford:
@imbstack so, this patch uses lib-url to generate the baseUrl.  I would rather that it be used to generate the entire url, but that would require a significantly larger patch and I don't know when I'd have time to get to it.  This would get us rootUrl support with the python client today, with the caveat that we don't switch to some format like "https://example.com/endpoint/path/v1/servicename"... which seems low risk.

jhford:
Btw, I've tested 2.7 and 3.7 locally, but don't have 3.5 or 3.6

taskcluster/client.py
```diff
@@ -67,6 +75,17 @@ def __init__(self, options=None, session=None):
...
+            s = 'specifying a rootUrl requires a version in the API reference'
+            raise exceptions.TaskclusterFailure(s)
+        elif rootUrl:
+            name = o.get('name', urllib.parse.urlparse(baseUrl).netloc.split('.', 1)[0])
+            o['baseUrl'] = liburl.api(rootUrl, name, 'v' + o['version'], '')
```
djmitche: As @imbstack indicated, this is wrong -- `o['version'] is the version of the reference format.  The [schema](https://github.com/taskcluster/taskcluster-references/blob/master/schemas/api-reference-v0.yml#L81) enforces that it is always `0`.

djmitche: For example, it's `0` for [secrets](http://references.taskcluster.net/secrets/v1/api.json)

jhford: How should we figure out which API version?  Do we default to a static `1` until we figure out how to specify that data?  I didn't see anything like apiVersion in the schema.

djmitche: It's in the manifest, not in the API schema.  So during parsing of the manifest we'll need to extract it.

jhford: But we don't use manifests everywhere.  An API reference is a description of an interface to a service, the version of that interface is an important part of that description.

djmitche: It is [there](https://github.com/taskcluster/taskcluster-lib-api/blob/master/src/schemas/api-reference.json), too.  Nobody has figured out how to synchronize those yet.

djmitche: https://bugzilla.mozilla.org/show_bug.cgi?id=1145220 is part of figuring that out.

PR #109, opened 2018-06-28
taskcluster.async -> taskcluster.aio to fix py37

escapewindow:
`async` is now a keyword in py37, so `import taskcluster.async` breaks.
This is a breaking change.


escapewindow:
Thanks! Unbitrotted and pushed an attempt at adding 3.7 tests. We have to use 3.7-dev until https://github.com/travis-ci/travis-ci/issues/9815 is resolved.

jhford:
According to https://github.com/travis-ci/travis-ci/issues/9831#issuecomment-401757065, we need to use xenial to get the release 3.7?  Either way, it's not important to me, just that my digging suggested that 3.7-dev was a fairly old beta... not sure if it matters.

escapewindow:
Yeah, once there's an officially supported way I plan on going through and updating our 3.7-dev envs. If they close https://github.com/travis-ci/travis-ci/issues/9815 saying that's the new supported way, we can update, or we can update and switch if they say that was just a temporary workaround.

escapewindow:
@jhford , now that PyYAML is 3.7-compliant, it looks like a taskcluster-client.py release with this code is the only thing we need to get a green py37 scriptworker. Could we schedule a 4.0.0?

PR #112, opened 2018-07-31
Store `baseUrl` with a single trailing slash.

### LINKED ISSUE #111: `buildUrl` is building urls with double slashes
I think this is fallout from [the queue api baseUrl changing](https://github.com/taskcluster/taskcluster-client.py/commit/0ef0cdb7d0adab4b55cf8b404b061fdd90932b2b#diff-062fb8411914f59c459e908b061748b9R2068).

I'd love for us to fix it and add tests around this.

```
(swenv) akimoz: /src/tc/scriptworker (loglevel) [11:25:34]
10483$ python
Python 3.6.5 (default, May  9 2018, 17:51:49)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)] on darwin
>>> from taskcluster.aio import Queue
>>> q = Queue()
>>> q.buildUrl('getLatestArtifact', 'T6ShjYKnSWmiRzFw0QVCtA', 'public/foo')
'https://queue.taskcluster.net/v1//task/T6ShjYKnSWmiRzFw0QVCtA/artifacts/public%2Ffoo'
>>> from taskcluster import Queue as SyncQueue
>>> q2 = SyncQueue()
>>> q2.buildUrl('getLatestArtifact', 'T6ShjYKnSWmiRzFw0QVCtA', 'public/foo')
'https://queue.taskcluster.net/v1//task/T6ShjYKnSWmiRzFw0QVCtA/artifacts/public%2Ffoo'
>>>
```

Note the `v1//task`. This broke the tree when I rolled tc4 out.

escapewindow:
We currently assume `baseUrl` has no trailing slash, and programmatically add it in multiple places with varying logic.

This patch stores `baseUrl` with a single trailing slash, and stops adding slashes later. (If we're worried about people overriding baseUrl outside of init, we can add more calls to `singleTrailingSlash`, or we can just do something basic like `baseUrl + route`, like the comment says, instead.)

I believe [this api change](https://github.com/taskcluster/taskcluster-client.py/commit/0ef0cdb7d0adab4b55cf8b404b061fdd90932b2b#diff-062fb8411914f59c459e908b061748b9R2068) broke things. (This patch will fix it)

Also points at release py37 for travis.

Fixes #111.


escapewindow:
I think always cleaning the inputs has the least likelihood of breaking later on.

.travis.yml
```diff
@@ -10,8 +10,11 @@ matrix:
       env: TOXENV=py35
     - python: 3.6
       env: TOXENV=py36
-    - python: 3.7-dev
+    - python: 3.7
```
jhford: Cool, thanks for updating to a non-dev version

taskcluster/utils.py
```diff
@@ -405,3 +406,11 @@ def do_GET(self):
     while not creds[0]:
         server.handle_request()
     return creds[0]
+
+
+def singleTrailingSlash(url):
+    return '{}/'.format(url.rstrip('/'))
+
+
+def joinBaseUrlAndRoute(baseUrl, route):
```
jhford: If we did the singleTrailingSlash logic in this function, we would be safe from the concern about the baseurl changing after instantiation.

Also, since this is logic which is targetting client stuff specifically, what about making it available as BaseClient._joinBaseUrlAndRoute instead?

escapewindow: Sure. I'll go ahead and move the code.

escapewindow:
Looks like we need `urljoin` or tests fail :) gtk. How does that look @jhford ?

escapewindow:
Thank you!

Callek:
Huh @escapewindow @jhford  - Unless github is lying to me we never fully merged this in - https://github.com/taskcluster/taskcluster-client.py/pull/112/commits/a3423081ee006eeb784e01710379a968d8d5139e doesn't seem to have merged.

Callek:
err it did Lie to me :(

PR #114, opened 2018-09-21
Remove use of baseUrl

djmitche:
This removes the concept of a baseUrl, requiring a rootUrl instead.  The
serviceName and apiVersion are intuited from the references if not
supplied directly, using logic based on that in the JS
taskcluster-client.

This is a breaking change, so warrants a major version bump.

There was some discussion of retaining the ability to use baseUrls for purposes of injecting services into other clusters (@petemoore's example with purge-caches).  I think we can still support this, but let's not call it baseUrl.  I want to make a clean break with baseUrls, and raising exceptions when they are provided is a robust way to do that.  Also, I think the names "baseUrl" and "rootUrl" are too easy to confuse. 

So, let's call it something more specific to the purpose, like `overrideServiceUrlPrefix`, and implement it across the board in all of the clients (once the rootUrl work and new client generation support is done).

### DIFF
--- a/genCode.py
+++ b/genCode.py
@@ -77,12 +77,22 @@ def createStaticClient(name, api, genAsync=False):
 
     lines.append('    classOptions = {')
 
-    copiedOptions = ('baseUrl', 'exchangePrefix')
+    # apply a default for apiVersion; this can be removed when all services have
+    # been upgraded to provide apiVersion
+    if 'apiVersion' not in api:
+        api['apiVersion'] = 'v1'
+
+    copiedOptions = ('exchangePrefix',)
     for opt in copiedOptions:
         if api.get(opt):
-            lines.append('        "%s": "%s"' % (opt, api[opt]))
+            lines.append('        "%s": "%s",' % (opt, api[opt]))
+    lines.append('    }')
 
-    lines.extend(['    }', ''])
+    copiedProperties = ('serviceName', 'apiVersion')
+    for opt in copiedProperties:
+        if api.get(opt):
+            lines.append('    %s = %s' % (opt, repr(api[opt])))
+    lines.append('')
 
     # We need to build up some information about how the functions work
     functionInfo = {}
--- a/taskcluster/client.py
+++ b/taskcluster/client.py
@@ -21,7 +21,7 @@ import mohawk.bewit
 
 import taskcluster.exceptions as exceptions
 import taskcluster.utils as utils
-import taskcluster_urls as liburl
+import taskcluster_urls as liburls
 
 log = logging.getLogger(__name__)
 
@@ -54,16 +54,14 @@ class BaseClient(object):
     """
 
     def __init__(self, options=None, session=None):
+        if options and options.get('baseUrl'):
+            raise exceptions.TaskclusterFailure('baseUrl option is no longer allowed')
         o = copy.deepcopy(self.classOptions)
         o.update(_defaultConfig)
         if options:
-            # We can only check for mutual exclusivity of baseUrl and rootUrl at
-            # this point because by default the reference will have a baseUrl,
-            # and so if we check if the class or default baseUrl options, it'll
-            # always be set so it'll always fail when using the rootUrl option.
-            if options.get('baseUrl') and o.get('rootUrl'):
-                raise exceptions.TaskclusterFailure('baseUrl and rootUrl are mutually exclusive')
             o.update(options)
+        if not o.get('rootUrl'):
+            raise exceptions.TaskclusterFailure('rootUrl option is required')
 
         credentials = o.get('credentials')
         if credentials:
@@ -75,13 +73,6 @@ class BaseClient(object):
                     except:
                         s = '%s (%s) must be unicode encodable' % (x, credentials[x])
                         raise exceptions.TaskclusterAuthFailure(s)
-
-        baseUrl = o.get('baseUrl')
-        rootUrl = o.get('rootUrl')
-
-        if rootUrl:
-            name = o.get('name', urllib.parse.urlparse(baseUrl).netloc.split('.', 1)[0])
-            o['baseUrl'] = liburl.api(rootUrl, name, 'v%d' % o.get('apiVersion', 1), '')
 
         self.options = o
         if 'credentials' in o:
@@ -184,7 +175,7 @@ class BaseClient(object):
         route = self._subArgsInRoute(entry, routeParams)
         if query:
             route += '?' + urllib.parse.urlencode(query)
-        return self._joinBaseUrlAndRoute(route)
+        return liburls.api(self.options['rootUrl'], self.serviceName, self.apiVersion, route)
 
     def buildSignedUrl(self, methodName, *args, **kwargs):
         """ Build a signed URL.  This URL contains the credentials needed to access
@@ -253,11 +244,14 @@ class BaseClient(object):
             u.fragment,
         ))
 
-    def _joinBaseUrlAndRoute(self, route):
-        return urllib.parse.urljoin(
-            '{}/'.format(self.options['baseUrl'].rstrip('/')),
-            route.lstrip('/')
-        )
+    def _constructUrl(self, route):
+        """Construct a URL for the given route on this service, based on the
+        rootUrl"""
+        return liburls.api(
+            self.options['rootUrl'],
+            self.serviceName,
+            self.apiVersion,
+            route.rstrip('/'))
 
     def _makeApiCall(self, entry, *args, **kwargs):
         """ This function is used to dispatch calls to other functions
@@ -453,7 +447,7 @@ class BaseClient(object):
         the logic about doing failure retry and passes off the actual work
         of doing an HTTP request to another method."""
 
-        url = self._joinBaseUrlAndRoute(route)
+        url = self._constructUrl(route)
         log.debug('Full URL used is: %s', url)
 
         hawkExt = self.makeHawkExt()
@@ -561,6 +555,8 @@ class BaseClient(object):
 
 
 def createApiClient(name, api):
+    api = api['reference']
+
     attributes = dict(
         name=name,
         __doc__=api.get('description'),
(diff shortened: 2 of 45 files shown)


djmitche:
By the way, I see this as just finishing the work of [bug 1428421](https://bugzilla.mozilla.org/show_bug.cgi?id=1428421) / #106 since you (@jhford) mentioned you didn't have time to get to it.  Not trying to step on any toes or anything!

test/base/__init__.py
```diff
@@ -15,6 +15,12 @@
     log.addHandler(logging.StreamHandler())
 
 
+TEST_ROOT_URL = 'https://tc-tests.localhost'
```
jhford: I see that this URL is only used to test parsing and isn't resolved.  Since that's the case, could we instead make it `https://tc-tests.example.com`?  Subdomains of localhost work out of the box on linux, but on OS X they don't resolve.  We have a lot of contributors to this library who use OS X.  My concern is that this variable might be used to do tests against localhost in new tests and introduce variability between OS X and Linux.  Using a subdomain of example.com will work the same everywhere, so I think we should consider using it.

test/base/__init__.py
```diff
@@ -15,6 +15,12 @@
...
+
+# rootUrl of a real deployment (that needs no pre-configuration)
+REAL_ROOT_URL = 'https://taskcluster.net'
```
jhford: Other test helpers are attached to the TCTest class.  Let's keep it consistent and attach this to the TCTest class.

test/test_client.py
```diff
@@ -313,6 +328,8 @@ def raise_for_status(self):
 
 class TestMakeHttpRequest(ClientTest):
 
+    apiPath = 'https://tc-tests.localhost/api/fake/v1/test'
```
jhford: Since this is the same base url, maybe we can use the rootUrl from the TCTest value?  Either way, like in that file, we should use a sub-domain of `example.com` or some similar.

test/test_client.py
```diff
@@ -586,17 +597,17 @@ def test_no_args_no_kwargs(self):
 class TestBuildUrl(ClientTest):
 
     def test_build_url_positional(self):
-        expected = 'https://fake.taskcluster.net/v1/two_args_no_input/arg0/arg1'
+        expected = 'https://tc-tests.localhost/api/fake/v1/two_args_no_input/arg0/arg1'
```
jhford: I really liked the self.apiPath pattern used in the previous tests, maybe it's applicable here too?

genCode.py
```diff
@@ -77,10 +77,24 @@ def createStaticClient(name, api, genAsync=False):
...
     lines.append('    classOptions = {')
 
-    copiedOptions = ('baseUrl', 'exchangePrefix')
+    # if serviceName is not set, get it from the baseUrl, exchangePrefix, or manifest name
+    # this can be removed when all services have serviceName
+    if 'serviceName' not in api:
+        if 'baseUrl' in api:
+            api['serviceName'] = api['baseUrl'].split('/')[2].split('.')[0]
+        elif 'exchangePrefix' in api:
+            api['serviceName'] = api['exchangePrefix'].split('/')[1].replace('taskcluster-', '')
```
jhford: If the pattern is `exchange/<serviceName>/<version>`, should that be added into taskcluster-lib-urls?

test/test_client.py
```diff
@@ -696,21 +707,22 @@ def test_no_creds_needed(self):
...
-                             'https://auth.taskcluster.net/v1/clients/abc')
+                             'https://tc.my.com/api/auth/v1/clients/abc')
             self.failIf('Authorization' in request.headers)
             headers = {'content-type': 'application/json'}
             content = {"clientId": "abc"}
             return httmock.response(200, content, headers, None, 5, request)
 
         with httmock.HTTMock(auth_response):
-            client = subject.Auth({"credentials": {}})
+            client = subject.Auth({"rootUrl": "https://tc.my.com", "credentials": {}})
```
jhford: Since this domain does resolve

```
[jhford@jhford ~]$ curl -I tc.my.com
HTTP/1.1 301 Moved Permanently
<snip>
Location: https://tc.my.com/
<snip>
```

could we make it a subdomain of example.com?

djmitche: Looks like it does resolve!  Sure, we should probably use the testing rootUrl anyway.

djmitche:
> My main concerns are the way service name is guessed, as well as that it is guessed. Since we're thinking of implementing a baseUrl functional equivalent, I think we should do that in the same pass through as this. If we do, we don't need to even try to guess the service name.

I think these are two separate issues -- guessing the serviceName (where one of the guesses is based on baseUrl from the api definition); and removing baseUrl support.

We could eliminate the serviceName guessing entirely with a patch to aws-provisioner - it's the only service still not producing one.  Probably the easiest would be forking an old version of tc-lib-api and making a patch release.  I could work on that if it would make this more acceptable.

As for the baseUrl breakage -- I kind of want tests to fail, loudly.  Warnings don't really achieve that. Having made this change in the JS client, I found it *really* helpful in updating apps depending on the library.  Fundamentally, this *is* a breaking change, and I feel like continuing to honor a baseUrl (which almost certainly has `taskcluster.net` in it if used in non-test code) is just hiding that that breakage from users.  Anyway, and it's your client, so I'll honor what you decide.

One thing I don't understand in your comment is how we could remove baseUrl support once all services have serviceName in their API definition.  Wouldn't that be precisely the breaking change you want to avoid?  I've probably just misunderstood something.

genCode.py
```diff
@@ -77,10 +77,24 @@ def createStaticClient(name, api, genAsync=False):
 
     lines.append('    classOptions = {')
 
-    copiedOptions = ('baseUrl', 'exchangePrefix')
+    # if serviceName is not set, get it from the baseUrl, exchangePrefix, or manifest name
+    # this can be removed when all services have serviceName
+    if 'serviceName' not in api:
+        if 'baseUrl' in api:
+            api['serviceName'] = api['baseUrl'].split('/')[2].split('.')[0]
```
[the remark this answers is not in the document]
djmitche: This could wouldn't ever get a bsaeUrl like that, because it's only looking at baseUrls in API definitions (not in constructor options), and only for old services where `serviceName` wasn't given (so, effectively, just aws-provisioner).  So we could just guess 'aws-provisioner' and actually be right every time :)

But as I said in my reply, we could fix aws-provisioner to provide serviceName, and then never need to apply this guessing at all.

genCode.py
```diff
@@ -77,10 +77,24 @@ def createStaticClient(name, api, genAsync=False):
 
     lines.append('    classOptions = {')
 
-    copiedOptions = ('baseUrl', 'exchangePrefix')
+    # if serviceName is not set, get it from the baseUrl, exchangePrefix, or manifest name
```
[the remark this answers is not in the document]
djmitche: It is in `createApiClient`.  I don't think it makes sense in the constructor, though - this is compatibility for handling older API definitions.  In other words, this is deriving default values for options from the API definition, rather than deriving option values from other options.

test/base/__init__.py
djmitche: It's the testRootUrl from tc-lib-urls, which is conveniently recognizable as such.  But, it doesn't really matter what it is.  I'll change it.

genCode.py
djmitche: It's not needed, because the full path is embedded in the exchanges.json -- rootUrl doesn't affect it.

djmitche:
Ah, on further reflection I think I'm starting to see where the disconnect is, here -- in current `master`, `options` is basically the data from the API reference file, with overrides provided by the user in the constructor arguments.  From that perspective, pre-processing any data in `genCode.py`/`createApiClient` doesn't make sense, and referencing `baseUrl` in that pre-processing, while not allowing it in the constructor arguments, doesn't make any sense.  Is that close to your perspective?  It does explain my confusion in the last paragraph ("Wouldn't that be..").

If so, why not change that model so that the two are distinct?
 * API reference: serviceName, apiVersion, exchangePrefix
 * Constructor options: rootUrl, credentials

Here, `genCode.py` and `createApiClient` take an API reference and produce something ready to construct (either as code or a class object) that is specific to the serviceName/apiVersion.  The constructor takes constructor options regarding how to connect to a specific instance of that service.  Then `baseUrl` appears nowhere (assuming we modify aws-provisioner) or only as an implementation detail of handling old-fashioned API definitions in `genCode.py` / `createApiClient`.  We can safely raise an exception if it's provided to the constructor.

This model change is sort of what I've done in this PR, but I could tighten it up a bit and make it clearer.

genCode.py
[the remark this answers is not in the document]
jhford: I think that a point release to lib-api for aws-provisioner is the way to go then.

genCode.py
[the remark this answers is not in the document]
jhford: Yes, it's compatiblity code, but the same code is being specified in here in the code generation and again in createApiClient.  We have other initialization code for clients in the constructor, and by moving this initialization code into the constructor, we don't need to have it duplicated in two places.

But if we remove the guessing of service name with a point release to lib-api, almost all of this code goes away, especially if there's a constant apiVersion of 'v1' added to that patch.

test/base/__init__.py
jhford: we probably should consider changing it there as well, since it has different behaviour based on the OS that the tests run on.

djmitche:
Yeah, that sounds good.

taskcluster/client.py
```diff
@@ -184,7 +179,7 @@ def buildUrl(self, methodName, *args, **kwargs):
         route = self._subArgsInRoute(entry, routeParams)
         if query:
             route += '?' + urllib.parse.urlencode(query)
-        return self._joinBaseUrlAndRoute(route)
+        return liburls.api(self.rootUrl, self.serviceName, self.apiVersion, route)
```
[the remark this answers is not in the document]
djmitche: I moved this back into a method so that functionality can be added more easily later.

test/base/__init__.py
djmitche: https://github.com/taskcluster/taskcluster-lib-urls/pull/16

djmitche:
OK, updated with the suggested changes.

djmitche:
(test scripts fixed up as indicated)

djmitche:
@jhford it doesn't look like I have permission to release this to pypi.  Can you merge and do the release?

djmitche:
Thanks!

PR #125, opened 2019-03-01
Stop linking to schemas (closes #120)

### LINKED ISSUE #120: Links in README.md are down
For example, https://github.com/taskcluster/taskcluster-client.py#get-task-definition links "Output schema" to https://github.com/taskcluster/taskcluster-client.py/blob/master/v1/task.json# -- which doesn't exist.

jhford:
r? @escapewindow


escapewindow:
I'm guessing we're removing these because of at least one/some of the following reasons?

- r14y breaks some assumptions about where the docs are
- tc.py hasn't been updated for a while, and the apis have changed enough that the links were old
- other ?

I think this approach works for removing the links, although strings like `This exchange has outputs` are a bit awkward.

I'm wondering if the best approach is removing the links, or fixing them in an r14y world.

djmitche:
> I'm guessing we're removing these because of at least one/some of the following reasons?
> 
>     * r14y breaks some assumptions about where the docs are
> 
>     * tc.py hasn't been updated for a while, and the apis have changed enough that the links were old
> 
>     * other ?
> 
> 
> I think this approach works for removing the links, although strings like `This exchange has outputs` are a bit awkward.
> 
> I'm wondering if the best approach is removing the links, or fixing them in an r14y world.

Issue #128, opened 2019-03-08
enable pyup and/or unpin dependencies


escapewindow:
Since the releng and treeherder teams regularly use pyup, the fact that

- tc.py pins its deps, and
- tc.py doesn't regularly update these deps,

causes problems downstream. We should either regularly update these deps and release, or unpin. The latter is easier for maintenance but may result in downstream bustage. The former is more maintenance, but would also pick up any new taskcluster api changes.

escapewindow:
Unpinned in #130.

PR #130, opened 2019-03-11
Unpin reqs in setup.py

escapewindow:
We currently have somewhat complex requirements in setup.py, to support py2-install, py3-install, py2-test, and py3-test. We could address this with a tool like pip-compile-multi, with `base.in`, `aio.in`, `test.in`, `test-aio.in` unpinned requirements files that get expanded with pinning and hashes for all upstream deps.

However, this is also complicated by the fact that we get different sets of hashes and upstream deps based on host arch and python minor version. Also, as a library, we want to keep our pinning minimal: rather than specify the latest known good version of an upstream dep, we want to allow for a range of known good versions, in case other deps pin a different version.

We could support a large matrix of dependency tests: the first expected-good version, the latest expected-good version, and the latest unknown-status version of each dependency, across python minor versions. This would be thorough, but potentially time consuming to set up.

I'm currently leaning towards unpinning, and running [cron?] tests regularly against the latest version of dependencies. If these automated tests find errors, we can pin the dep to `<first.bad.version` until we fix. This PR is potentially the first step.

I believe this would resolve both #87 and #128.

### DIFF
--- a/devDep.py
+++ b/devDep.py
@@ -21,3 +21,4 @@ except:
     pass
 
 pipMain(['install', '--upgrade'] + setup.tests_require + setup.install_requires)
+pipMain(['freeze'])
--- a/setup.py
+++ b/setup.py
@@ -10,28 +10,28 @@ import sys
 VERSION = '7.0.0'
 
 tests_require = [
-    'nose==1.3.7',
-    'nose-exclude==0.5.0',
-    'httmock==1.2.6',
-    'rednose==1.2.1',
-    'mock==1.0.1',
-    'setuptools-lint==0.3',
-    'flake8==2.5.0',
-    'psutil==2.1.3',
-    'hypothesis==3.6.1',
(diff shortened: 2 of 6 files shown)


escapewindow:
There have been no objections to [this email](https://groups.google.com/forum/#!topic/mozilla.tools.taskcluster/hv32oq_a2FI). @djmitche do you feel comfortable reviewing this? Should I flag someone from releng?

escapewindow:
True. I think the other half of this would be to cron some tests, maybe weekly? Is there a standard way of doing that in taskcluster-github, or should I hack something in hooks until there is one?

djmitche:
I believe ci-taskgraph will provide more gecko-like support for such things, but at the moment "manual" creation of a hook is the best approach.

escapewindow:
Created a hook in #131.

